PR #19, opened 2021-03-08
Setting intial theme using user preference

### LINKED ISSUE #6: Set light/dark theme based on user pref
When the theme is not explicitly set, use the [prefers-colors-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) media query to set the right values

JayaKrishnaNamburu:
fixes #6

### DIFF
--- a/sandpack-react/src/contexts/theme-context.tsx
+++ b/sandpack-react/src/contexts/theme-context.tsx
@@ -1,15 +1,8 @@
 import * as React from 'react';
 import { useClasser } from '@code-hike/classer';
 import { createThemeObject, codesandboxLightTheme } from '../themes';
-import {
-  SandpackPartialTheme,
-  SandpackPredefinedTheme,
-  SandpackTheme,
-} from '../types';
-import {
-  injectThemeStyleSheet,
-  getDarkModePreference,
-} from '../utils/dom-utils';
+import { SandpackTheme, SandpackThemeProp } from '../types';
+import { injectThemeStyleSheet } from '../utils/dom-utils';
 
 const SandpackThemeContext = React.createContext<{
   theme: SandpackTheme;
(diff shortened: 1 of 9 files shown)


sandpack-react/src/contexts/theme-context.tsx
```diff
@@ -19,13 +22,16 @@ const SandpackThemeContext = React.createContext<{
...
-  const { theme, id } = createThemeObject(props.theme);
+  let themeObj = createThemeObject(
+    getDarkModePreference() ? 'codesandbox-dark' : 'codesandbox-light'
+  );
   const c = useClasser('sp');
-
   // If theme is not explicitly set, don't inject any stylesheet
   if (props.theme) {
-    injectThemeStyleSheet(theme, id);
+    themeObj = createThemeObject(props.theme);
```
alexnm, line 31: if you make the changes above, you can keep the injectThemeStyleSheet behind the props.theme check

alexnm (changes requested):

JayaKrishnaNamburu:
sure alex, wll be making those changes 👍

sandpack-react/src/themes/index.ts
```diff
@@ -175,9 +176,18 @@ export const SANDPACK_THEMES: Record<SandpackPredefinedTheme, SandpackTheme> = {
   'monokai-pro': monokaiProTheme,
 };
 
-export const createThemeObject = (
-  inputTheme?: SandpackPredefinedTheme | SandpackPartialTheme
-) => {
+export const createThemeObject = (inputTheme?: SandpackThemeProp) => {
+  if (inputTheme === 'auto') {
+    const theme = getDarkModePreference()
```
alexnm, line 181: here you can also assign the id to `codesandbox-light` or `codesandbox-dark` so you don't generate a new hash for the theme. `auto` will still set a standard theme, the hash function is created for custom themes. Also, move this after the `if(inputTheme === undefined)`

alexnm (changes requested):
Looks great, thanks for the effort! Just a small change request and we're good to go

alexnm (approved):

Issue #23, opened 2021-03-15
Self hosting the bundler


vipulbhj:
Here the documentation mentions that we can self host the bundler, 
> I want to highlight that you can also host the bundler by yourself, all necessary files are in the sandpack folder.

It mentions the `sandpack` folder, which I am confused about slightly as to which folder are we talking about here and also, how would I go about it If I wanted to self host the bundler. If I understand correctly, Sandpack imports currently point to `...${version}_codesandbox.io...` and I haven't found any option to change this to some other url. 

Can you explain the process of self hosting bundler a little more please.

alexnm:
Yes. I will try to document this in the `sandpack-client` readme.

- the bundler is part of the codesandbox-client codebase: https://github.com/codesandbox/codesandbox-client
- you can create your instance after you install dependencies with `yarn` if you run `yarn build:sandpack` 
- this creates a `www` folder in the root folder (I know, not the best naming)
- that `www` folder is the `sandpack` folder that `sandpack-client` connects to on `{version}-sandpack.codesandbox.io`
- once you have this hosted on your end, you can pass `bundlerURL` when calling `new SandpackClient(iframe, sandboxInfo, { bundlerURL: 'https://your-hosted-version' } )`
- or, if you use `sandpack-react`, you can bundlerURL in the options of the `Sandpack` preset

vipulbhj:
Hey @alexnm 

Thank you so much for all your help :)

alexnm:
Sure, anytime.

If the steps work for you, feel free to submit a PR to the readme of `sandpack-client`, with a section after the `Why is the bundler hosted externally`. every contribution is much appreciated at this point, since we just started testing this project. 

And also, let us know what you build with it, we'd love to showcase some of the community projects around sandpack

vipulbhj:
Sounds good, I will keep you in the loop as to how our experiments go :)

vipulbhj:
Hey @alexnm ,  I am getting this error while running the build command as you had mentioned.

```javascript
➜ yarn build:sandpack

yarn run v1.22.10
$ lerna run build:sandpack-sandbox --scope app --stream
lerna notice cli v3.16.4
lerna info versioning independent
lerna info filter [ 'app' ]
lerna info Executing command in 1 package: "npm run build:sandpack-sandbox"
app: > app@0.1.1 build:sandpack-sandbox /Users/vipulbardwaj/Work/POCs/codesandbox-client/packages/app
[...]

PR #25, opened 2021-03-18
Added Docs for self hosting the bundler

vipulbhj:
Content is taken from #23

### DIFF
--- a/sandpack-client/README.md
+++ b/sandpack-client/README.md
@@ -266,8 +266,10 @@ If you want to host the bundler yourself, you will need to do a few things.
 * `yarn build:deps` to build some of the packages lerna needs for internal links.
 * create your instance of sandpack with `yarn build:sandpack`.
 ```
-This creates a www folder in the root folder that www folder is the sandpack folder that sandpack-client connects to on {version}-sandpack.codesandbox.io
-once you have this hosted on your end, you can pass bundlerURL when calling new SandpackClient(iframe, sandboxInfo, { bundlerURL: 'https://your-hosted-version' } )
+This creates a www folder in the root folder that www folder is the sandpack folder.
+The sandpack-client connects to on {version}-sandpack.codesandbox.io, 
+Once you have this hosted on your end you can pass bundlerURL when calling
+new SandpackClient(iframe, sandboxInfo, { bundlerURL: 'https://your-hosted-version' } )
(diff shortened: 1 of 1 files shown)


sandpack-client/README.md
```diff
@@ -257,6 +257,19 @@ worker externally.
 > I want to highlight that you can also host the bundler by yourself, all
 > necessary files are in the `sandpack` folder.
 
+### Self Hosting Bundler
```
alexnm, line 260: make it a second level heading here

sandpack-client/README.md
```diff
@@ -257,6 +257,19 @@ worker externally.
...
+
+If you want to host the bundler yourself, you will need to do a few things.
+
+* The bundler is part of the codesandbox-client codebase: https://github.com/codesandbox/codesandbox-client
+* Clone the codesandbox-client and install the dependencies in the root folder (`yarn install`).
+* `yarn build:deps` to build some of the packages lerna needs for internal links.
+* create your instance of sandpack with `yarn build:sandpack`.
+
+*This creates a www folder in the root folder that www folder is the sandpack folder that sandpack-client connects to on {version}-sandpack.codesandbox.io
```
alexnm, line 269: I would put www and {version}-sandpack.codesandbox.io between back ticks for better format. Also, it feels like a space is missing here after the `*` for the bullet list

sandpack-client/README.md
```diff
@@ -257,6 +257,19 @@ worker externally.
...
+once you have this hosted on your end, you can pass bundlerURL when calling new SandpackClient(iframe, sandboxInfo, { bundlerURL: 'https://your-hosted-version' } )
```
alexnm, line 270: can you try adding the code snippet in a code block formatted between triple-backticks? It becomes more readable. also, you should format this with prettier if it's possible, it will rearrange the readme content to a max of 80 characters per line I think. don't have prettier on pre-commit yet, but there's a task for that on the issues list

alexnm (changes requested):
Great, thanks for the contribution! Left a few aesthetic related comments, when you have time to check it out

sandpack-client/README.md
vipulbhj: Sure

vipulbhj:
Changed things according to feedback

alexnm (approved):
I made the whole backticks thing a bit confusing, I will merge this and make an update for the format, thanks!

PR #36, opened 2021-03-30
error screen details and small fixes

### LINKED ISSUE #14: Runtime errors cause flickering and don't show enough info
1. Runtime errors do not have sufficient info (line/row, stack trace)
2. Preview flickers because of the runtime error happening after the compile was done without any errors (and the error overlay disappears) There are a couple of options here, but they involve some work in the bundler which is not straight-forward.

### LINKED ISSUE #15: Option to disable the console.clear the bundler does
Can mess up the DX, especially if preserve log does not work with console.clear (like in Firefox)

### LINKED ISSUE #16: Toggle features inside the bundler based on the flags
sandpack-client now sends 3 flags on compile:
- showOpenInCodeSandbox
- showErrorScreen
- showLoadingScreen

Only the open in csb feature is implemented

### LINKED ISSUE #32: Styles and fonts in the head tag of index.html are not reflecting for react.
I am trying to render the same project in code-sandbox and sandpack. Looks likes anything in the head of `index.html` is not reflecting in the render output.

I am not sure, if the error is on the side of bundler or config of the project.

Here is the link for codesandbox --> https://codesandbox.io/s/1cugh?file=/public/index.html
Here is the sandpack implementation to render the same project --> https://repl.teleporthq.io/project/9ea7c3fc-46b3-4dcb-bb01-01bdecdd99f1

alexnm:
closes #32 - head tag replacing the original html head in the bundler on the first load
closes #14 - more information shown on runtime errors
closes #15 - option set to false in sandpack-client, can later be overwritten
closes #16 - inside the bundler

### DIFF
--- a/sandpack-client/src/client.ts
+++ b/sandpack-client/src/client.ts
@@ -12,14 +12,18 @@ import type {
   Dependencies,
   SandpackBundlerFiles,
   BundlerState,
-  ModuleError,
   Modules,
   ClientStatus,
   UnsubscribeFunction,
   SandpackMessage,
   ListenerFunction,
+  SandpackError,
 } from "./types";
-import { createPackageJSON, addPackageJSONIfNeeded } from "./utils";
+import {
+  createPackageJSON,
+  addPackageJSONIfNeeded,
+  extractErrorDetails,
+} from "./utils";
 
 export interface ClientOptions {
   /**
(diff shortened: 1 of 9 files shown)


danilowoz (commented):
I still don't know much about the architecture decision, but the code is very clear and looks good to me.

PR #42, opened 2021-04-20
Sandpack Playground

DeMoorJasper:
I've added a package that is very useful for debugging bundler issues without having to start storybook (which is fairly heavy and slow)

The only thing that really remains to make this the perfect debug tool is having a way to provide a sandbox id and loading that sandbox instead of hardcoding file content. Not entirely sure yet how that would work, might look into that after fixing some issues with the cat example

Screen Shot 2021-04-20 at 12 53 22

### DIFF
--- a/sandpack-playground/src/App.tsx
+++ b/sandpack-playground/src/App.tsx
@@ -1,19 +1,30 @@
 import { Sandpack } from "@codesandbox/sandpack-react";
 import React from "react";
 
+const reactCode = `export default function App() {
+  return (
+    <div>
+      <h1>Hello World</h1>
+    </div>
+  )
+}
+`;
+
 export const App: React.FC = () => {
   return (
-    <Sandpack
-      files={{
-        "/App.js": "",
-      }}
-      options={{
-        autorun: false,
-        showTabs: true,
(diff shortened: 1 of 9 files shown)


sandpack-playground/src/App.tsx
```diff
@@ -0,0 +1,27 @@
...
+      <Sandpack
+        files={{
+          "/App.js": reactCode,
+        }}
+        options={{
+          autorun: true,
+          showTabs: true,
+          showLineNumbers: true,
+          showNavigator: true,
+          editorHeight: "calc(100vh - 5px)",
```
DeMoorJasper, line 21: This is a bit of a hack to get it to full-height, not really sure how else to do this as the editor doesn't scale with the outer container even when providing 100% height.

alexnm: I can update the usage here once we merge this, there's a prop through which you can pass css classes that attach to sandpack classes

alexnm:
loading a sandbox by id is a feature we want to implement in sandpack as well in the next period. but so far, sandpack has been completely independent from the codesandbox backend, so we have to see which is the best way to make the integration

PR #43, opened 2021-04-20
Patch/0.0.4

### LINKED ISSUE #5: Loading animation timeout
The preview overlay animation could have a timeout after which it disappears to let the users know something is wrong. There might be scenarios we're not accounting for, in which the messages are not dispatched by the bundler so the overlay covers the potential error indefinitely.

### LINKED ISSUE #39: updateCurrentFile adds an extra character to new code
`updateCurrentFile` method from `sandpack` object adds an  extra } character to the previewed code. 
So the preview window shows this source code:
````typescript
import React, { FC } from "react";

export const Main: FC = () => {
    const buttonClick = () => {
        window.parent.postMessage({ type: 'click', message: { element: 'button' }}, '*')
    }

    return <button onClick={buttonClick}>Trigger event</button>
}}
````
Reproduction repository: [https://github.com/nkovacic/sandpack-examples](https://github.com/nkovacic/sandpack-examples)

Full example:

```` jsx
import React, { FC } from 'react';
import './App.css';

import {
  SandpackCodeEditor,
  SandpackLayout,
  SandpackPreview,
  SandpackProvider,
  SandpackSetup
} from '@codesandbox/sandpack-react';

import '@codesandbox/sandpack-react/dist/index.css';

const mainFile = `import React, { FC } from "react";

export const Main: FC = () => {
    const buttonClick = () => {
        window.parent.postMessage({ type: 'click', message: { element: 'button' }}, '*')
    }

    return <button onClick={buttonClick}>Trigger event</button>
}`;

const reactTypescriptCustomSetup: SandpackSetup = {
  entry: '/src/index.tsx',
  main: '/src/main.tsx',
  dependencies: {
    react: 'latest',
    'react-dom': 'latest',
    'react-scripts': '4.0.0',
  },
  files: {
    './tsconfig.json': {
      code: `{
  "include": [
    "./src/**/*"
  ],
  "compilerOptions": {
    "strict": true,
    "esModuleInterop": true,
    "lib": [
      "dom",
      "es2015"
    ],
    "jsx": "react"
  }
  }`,
      hidden: true,
    },
    '/public/index.html': {
      code: `<!DOCTYPE html>
  <html lang="en">
  <head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <title>Document</title>
  </head>
  <body>
  <div id="root"></div>
  </body>
  </html>`,
      hidden: true,
    },

    '/src/index.tsx': {
      code: `import * as React from "react";
  import { render } from "react-dom";
  import { Main } from "./main";
  const rootElement = document.getElementById("root");
  render(<Main/>, rootElement);
          `,
      hidden: true,
    },

    '/src/main.tsx': {
      code: mainFile,
    },
  },
};

const SandpackListener: FC = () => {
    const { sandpack: { updateCurrentFile} } = useSandpack()

    useEffect(() => {
        const listener = (event: MessageEvent) => {
            if (event.data?.type === 'click') {
                const newMainFile = mainFile.replace('Trigger', 'Do not trigger');

                console.log(newMainFile);

                updateCurrentFile(newMainFile);
            }
        }

        window.addEventListener("message", listener, false);

        return () => window.removeEventListener('message', listener)
    }, []);

    return null;
}

const App: FC = () => {
  return (
    <div className="App">
      <SandpackProvider customSetup={reactTypescriptCustomSetup}>
        <SandpackLayout>
          <SandpackCodeEditor />
          <SandpackListener />
          <SandpackPreview
            showOpenInCodeSandbox={false}
            showRefreshButton={false}
          />
        </SandpackLayout>
      </SandpackProvider>
    </div>
  );
};

export default App;
````

### PR DESCRIPTION
**Pull request by alexnm**
Closes #5 
Closes #39

### DIFF
--- a/sandpack-react/src/hooks/useLoadingOverlayState.ts
+++ b/sandpack-react/src/hooks/useLoadingOverlayState.ts
@@ -2,7 +2,11 @@ import * as React from "react";
 
 import { useSandpack } from "./useSandpack";
 
-export type LoadingOverlayState = "visible" | "fading" | "hidden";
+export type LoadingOverlayState = "visible" | "fading" | "hidden" | "timeout";
+
+const BUNDLER_TIMEOUT = 30000; // 30 seconds timeout for the bundler to respond.
+const FADE_DELAY = 1000; // 1 second delay one initial load, only relevant if the loading overlay is visible.
+const FADE_ANIMATION_DURATION = 500; // 500 ms fade animation
 
 export const useLoadingOverlayState = (): LoadingOverlayState => {
   const { sandpack, listen } = useSandpack();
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: sandpack-react/src/hooks/useLoadingOverlayState.ts [hunk 1]
```diff
@@ -2,7 +2,11 @@ import * as React from "react";
 
 import { useSandpack } from "./useSandpack";
 
-export type LoadingOverlayState = "visible" | "fading" | "hidden";
+export type LoadingOverlayState = "visible" | "fading" | "hidden" | "timeout";
+
+const BUNDLER_TIMEOUT = 300; // 30 seconds timeout for the bundler to respond.
```
DeMoorJasper, line 7: Shouldn't the comment be `300 ms timeout for the bundler to respond.`?
Also maybe it's a good idea to increase this slightly (maybe 1 second) to account for slower mobile devices and places with slow internet.

alexnm: ah, thanks for spotting that, I ran a test for the second commit and forgot to update back to 30000. I left at 30 seconds for now to be sure there's no latency involved. I've seen scenarios when multiple sandpacks need more than 10 seconds to load on slow connections. So, 30 seconds is safe for a timeout. This should be an extreme case nevertheless.

PR #46, opened 2021-04-21
fix: loading timeout only counted after the client is instantiated

alexnm:
Had to lift the timeout calculation to the sandpack context because the hook can be mounted before the client is instantiated (eg: when not using the autorun option)

### DIFF
--- a/sandpack-react/src/contexts/sandpackContext.tsx
+++ b/sandpack-react/src/contexts/sandpackContext.tsx
@@ -25,6 +25,7 @@ import { getSandpackStateFromProps } from "../utils/sandpackUtils";
 import { generateRandomId } from "../utils/stringUtils";
 
 const Sandpack = React.createContext<SandpackContext | null>(null);
+const BUNDLER_TIMEOUT = 30000; // 30 seconds timeout for the bundler to respond.
 
 export interface SandpackProviderState {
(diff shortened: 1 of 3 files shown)


sandpack-react/src/contexts/sandpackContext.tsx
```diff
@@ -294,6 +300,14 @@ class SandpackProvider extends React.PureComponent<
     // Clear the queued listeners after they were registered
     this.queuedListeners = {};
     this.setState({ sandpackStatus: "running" });
+
+    if (this.timeoutHook) {
```
alexnm, line 304: this is more like a fail safe, in case the run function is called multiple times for whatever reason

sandpack-react/src/types.ts
```diff
@@ -41,7 +41,7 @@ export interface SandpackState {
   loadingScreenRegisteredRef: React.MutableRefObject<boolean>;
 }
 
-export type SandpackStatus = "initial" | "idle" | "running";
+export type SandpackStatus = "initial" | "idle" | "running" | "timeout";
```
alexnm, line 44: timeout is a bit more explicit now inside the state of sandpack

DeMoorJasper (approved):

PR #48, opened 2021-04-22
fix: intersection observer threshold

alexnm:
this one is short :)


alexnm:
great job Alex :) (testing notification in frontapp)

PR #73, opened 2021-07-02
fix(listener): add global queue of listener to be registered

danilowoz:


sandpack-react/src/contexts/sandpackContext.tsx
```diff
@@ -338,7 +366,6 @@ class SandpackProvider extends React.PureComponent<
     if (client) {
       client.cleanup();
       delete this.clients[clientId];
-      delete this.unsubscribeQueuedListeners[clientId];
```
danilowoz, line 341: This was causing the main error when unmounting a client. So, I think we can remove this at all since unsubscription functions need to handle the entire list before cleaning themselves from this list.

PR #74, opened 2021-07-02
Listener is missing to register emit

### LINKED ISSUE #71: Listener is missing to register/emit events in some cases.
It was quite tricky to point down exactly when it's happening. But at times, the `listener` from `useSandpack` is failing to listen for events emitted by the sandpack bundler.
https://codesandbox.io/s/eloquent-mccarthy-qr09r?file=/src/App.tsx

STR
- Open the console and we should be able to see the logs.
- Refresh the browser or `csb` preview. And in some instances the listener won't log the events.

https://user-images.githubusercontent.com/11075561/123729554-c96bde00-d8b2-11eb-8301-8de480e06884.mov

danilowoz:
Basically, the main problem with not registering the listener was because the client was created after the listening has been added. So, I created a new object in the `queuedListeners` called `global`, to take care of all the events that must be added to clients, regardless of when. So I guess we cover the following cases:
 - Set a listener, but the client hasn't been created yet;
 - Set a listener, but the client has already been created;
 - A client already exists, set a new listener and then one more client has been created;

Closes #71

### DIFF
--- a/sandpack-react/src/contexts/sandpackContext.tsx
+++ b/sandpack-react/src/contexts/sandpackContext.tsx
@@ -446,24 +446,25 @@ class SandpackProvider extends React.PureComponent<
 
       // Add to the current clients
       const clients = Object.values(this.clients);
-      const currentClientListeners = clients.map((client) =>
+      const currentClientUnsubscribeListeners = clients.map((client) =>
         client.listen(listener)
       );
-      const currentClientListenersUnsubscribe = () => {
-        currentClientListeners.forEach((unsubscribe) => unsubscribe());
-      };
 
       const unsubscribeListener = () => {
         const unsubscribeQueuedClients = Object.values(
           this.unsubscribeQueuedListeners
         );
 
+        // Unsubscribing all listener registered
         unsubscribeQueuedClients.forEach((listenerOfClient) => {
           const listenerFunctions = Object.values(listenerOfClient);
           listenerFunctions.forEach((unsubscribe) => unsubscribe());
         });
 
-        currentClientListenersUnsubscribe();
+        // Unsubscribing from the clients already created
+        currentClientUnsubscribeListeners.forEach((unsubscribe) =>
+          unsubscribe()
+        );
       };
 
       return unsubscribeListener;
--- a/sandpack-react/src/contexts/sandpackContext.tsx
+++ b/sandpack-react/src/contexts/sandpackContext.tsx
@@ -113,7 +113,18 @@ class SandpackProvider extends React.PureComponent<
       renderHiddenIframe: false,
     };
 
-    this.queuedListeners = {};
+    /**
+     * List of functions to be registered in the client, once it has been created
+     *
+     * Use cases:
+     * - Set a listener, but the client hasn't been created yet;
+     * - Set a listener, but the client has already been created;
+     * - A client already exists, set a new listener and then one more client has been created;
+     */
+    this.queuedListeners = { global: {} };
+    /**
+     * Global list of unsubscribe function for the listeners
+     */
     this.unsubscribeQueuedListeners = {};
     this.preregisteredIframes = {};
     this.clients = {};
@@ -301,8 +312,10 @@ class SandpackProvider extends React.PureComponent<
       }, BUNDLER_TIMEOUT);
     }
 
+    /**
+     * Register any potential listeners that subscribed before sandpack ran
+     */
     if (this.queuedListeners[clientId]) {
-      // Register any potential listeners that subscribed before sandpack ran
       Object.keys(this.queuedListeners[clientId]).forEach((listenerId) => {
         const listener = this.queuedListeners[clientId][listenerId];
         const unsubscribe = client.listen(listener) as () => void;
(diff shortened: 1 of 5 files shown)


sandpack-react/src/contexts/sandpackContext.tsx
```diff
@@ -338,7 +366,6 @@ class SandpackProvider extends React.PureComponent<
     if (client) {
       client.cleanup();
       delete this.clients[clientId];
-      delete this.unsubscribeQueuedListeners[clientId];
```
danilowoz, line 341: This was causing the main error when unmounting a client. So, I think we can remove this at all since unsubscription functions need to handle the entire list before cleaning themselves from this list.

sandpack-react/src/contexts/sandpackContext.tsx
```diff
@@ -313,6 +326,21 @@ class SandpackProvider extends React.PureComponent<
       this.queuedListeners[clientId] = {};
     }
 
+    /**
+     * Register global listeners
+     */
+    const globalListeners = Object.entries(this.queuedListeners.global);
```
alexnm, line 332: This is a good catch, and I like the approach

sandpack-react/src/contexts/sandpackContext.tsx
```diff
@@ -394,7 +423,8 @@ class SandpackProvider extends React.PureComponent<
           this.unsubscribeQueuedListeners[clientId] || {};
 
         this.queuedListeners[clientId][listenerId] = listener;
-        return () => {
+
+        const unsubscribeListener = () => {
```
alexnm, line 427: Question: Is this useful for debugging or why is the change needed?

sandpack-react/src/contexts/sandpackContext.tsx
alexnm: Yes, this is a bit over-optimized, the important thing is that the unsubscribe functions get called. `this.unsubscribeQueuedListeners[clientId]` would not be used after this if block executes

sandpack-react/src/contexts/sandpackContext.tsx
```diff
@@ -406,12 +436,37 @@ class SandpackProvider extends React.PureComponent<
...
       }
     } else {
-      const unsubs = Object.values(this.clients).map((client) =>
+      // Push to the **global** queue
+      const listenerId = generateRandomId();
+      this.queuedListeners.global[listenerId] = listener;
+
+      // Add to the current clients
+      const clients = Object.values(this.clients);
+      const currentClientListeners = clients.map((client) =>
```
alexnm, line 449: nitpick: I would rename this to currentClientUnsubscribeListeners, for clarity

sandpack-react/src/contexts/sandpackContext.tsx
```diff
@@ -406,12 +436,37 @@ class SandpackProvider extends React.PureComponent<
...
+      const currentClientListenersUnsubscribe = () => {
+        currentClientListeners.forEach((unsubscribe) => unsubscribe());
+      };
+
+      const unsubscribeListener = () => {
+        const unsubscribeQueuedClients = Object.values(
+          this.unsubscribeQueuedListeners
+        );
+
+        unsubscribeQueuedClients.forEach((listenerOfClient) => {
```
alexnm, line 461: Can you leave a comment for this block as well? I believe it is unsubscribing the listeners which were initially queued globally and were added to each unsubscribeQueuedListeners[clientId], right?

sandpack-react/src/contexts/sandpackContext.tsx
```diff
@@ -406,12 +436,37 @@ class SandpackProvider extends React.PureComponent<
...
+      const listenerId = generateRandomId();
+      this.queuedListeners.global[listenerId] = listener;
+
+      // Add to the current clients
+      const clients = Object.values(this.clients);
+      const currentClientListeners = clients.map((client) =>
         client.listen(listener)
       );
-      return () => unsubs.forEach((unsub) => unsub());
+      const currentClientListenersUnsubscribe = () => {
```
alexnm, line 452: nitpick: because there are already a bunch of functions/variables here with similar names, I would keep this as an anonymous function at line 466, because it is similar with the other foreach that unsubscribes from each "future" client. Also, `current` might be misleading in the name, since timeline is super complex in this function :)

alexnm (approved):
Really great job @danilowoz, this has to be the most complicated flow in the entire library and you found the issue with ease.

sandpack-react/src/contexts/sandpackContext.tsx
danilowoz: Yeah, I wouldn't worry about this extra optimization, because the object will be empty at the end of unsubscribing listener process.

sandpack-react/src/contexts/sandpackContext.tsx
danilowoz: I did it just to make the code clearer, now each conditional returns the same kind of function and which makes it easy to read long complex parts of code.

alexnm (approved):

PR #75, opened 2021-07-08
Close file action in the context and from files tab

JayaKrishnaNamburu:
Heya, i am just adding a small `closeFile` feature. Please let me know if this would e good to have in the react components. I will work little more on the UX and related stuff :)
Just added a basic stuff for the feedback

https://user-images.githubusercontent.com/11075561/124879150-8ca69200-dfea-11eb-93be-c2de7919868d.mov


alexnm:
Interesting, but gets you to a point where you cannot reopen the files, so I wouldn't introduce this as a standard feature in sandpack, but maybe it can be a different custom components (closable tabs) which can go well with the file explorer component (which is not really polished)

JayaKrishnaNamburu:
Yes, at the moment it can stay as a wrapper as a separate comp. And in future as a `flag` thing or auto enable when `FilesExplorer` is used. And how about a couple of utility functions, `closeFile`, `deleteFile`, Adding a file can work with `updateFile` too.

JayaKrishnaNamburu:
Hey @alexnm, made the changes. Now made them available via flag, which can be passed via `SandpackCodeEditor` or `Sandpack` itself. And added a extra utility function called `deleteFile` and exposed in context. Which helps in deleting files in sandpack from outside 😄. Helps in case where we are overwriting with a custom project setup. But the default files like `App.js` are still sticking around. So, this can help in deleting them if not needed by the users.

alexnm:
@danilowoz would you have time to review this?

sandpack-react/src/styles/index.css
```diff
@@ -254,6 +253,22 @@
   box-shadow: inset 0 0 0 2px var(--sp-colors-accent);
 }
 
+.sp-close-button {
+  padding: 0px var(--sp-space-1) 2px var(--sp-space-1);
+  border-radius: var(--sp-border-radius);
+  margin-left: var(--sp-space-1);
+  width: 20px;
+  visibility: hidden;
```
danilowoz, line 261: Consider adding a transition between states, it would look better 👍

sandpack-react/src/components/CodeEditor/index.tsx
```diff
@@ -25,11 +26,12 @@ export const SandpackCodeEditor: React.FC<CodeEditorProps> = ({
   showLineNumbers = false,
   showRunButton = true,
   wrapContent = false,
+  closableTabs = false,
 }) => {
   const { sandpack } = useSandpack();
   const { code, updateCode } = useActiveCode();
   const { activePath, status, editorState } = sandpack;
-  const shouldShowTabs = showTabs ?? sandpack.openPaths.length > 1;
+  const shouldShowTabs = showTabs ?? sandpack.openPaths.length > 0;
```
danilowoz, line 34: Is there any reason why you have changed this? I think it doesn't belong to the closable tabs.

sandpack-react/src/components/FileTabs/FileTabs.stories.tsx
```diff
@@ -27,6 +27,23 @@ export const Component: React.FC = () => (
   </SandpackProvider>
 );
 
+export const WithClosableTabs: React.FC = () => (
```
danilowoz, line 30: Great!

sandpack-react/src/components/FileTabs/index.tsx
```diff
@@ -29,6 +46,11 @@ export const FileTabs: React.FC = () => {
             type="button"
           >
             {getFileName(filePath)}
+            {closableTabs && openPaths.length > 0 ? (
```
danilowoz, line 49: In order to improve the user experience, I wouldn't render the close button when it's a single child. Based on your screen recording and internal test, actually you can close it where there is only one tab item. Does it make sense?

sandpack-react/src/contexts/sandpackContext.tsx
```diff
@@ -166,24 +166,26 @@ class SandpackProvider extends React.PureComponent<
...
     const newFiles = {
       ...files,
       [path]: { code: newCode },
     };
 
-    this.setState({ files: newFiles });
+    this.setState({ files: newFiles }, this.updateClients);
+  };
 
+  updateClients = (): void => {
```
danilowoz, line 178: Looks good!

sandpack-react/src/contexts/sandpackContext.tsx
```diff
@@ -389,6 +391,45 @@ class SandpackProvider extends React.PureComponent<
...
+  closeFile = (path: string): void => {
+    if (this.state.openPaths.length === 1) {
+      return;
+    }
+
+    this.setState(({ openPaths, activePath }) => {
+      const newPaths = openPaths.filter((openPath) => openPath !== path);
+
+      return {
+        activePath: path === activePath ? newPaths[0] : activePath,
```
danilowoz, line 403: Instead of setting the first item as active, I'd set the item before that was just closed, as exactly how any tabs (browser, vscode, so on) works. 

For example:
https://user-images.githubusercontent.com/4838076/125816787-f8a10cd3-b588-4e1b-891d-fdfaceb97cfc.mov

sandpack-react/src/contexts/sandpackContext.tsx
```diff
@@ -389,6 +391,45 @@ class SandpackProvider extends React.PureComponent<
...
+        openPaths: newPaths,
+        editorState: "dirty",
+      };
+    });
+  };
+
+  deleteFile = (path: string): void => {
```
danilowoz, line 410: I didn't get where you're using this method, what it's the purpose of it?

sandpack-react/src/types.ts
```diff
@@ -54,7 +57,12 @@ export interface SandpackState {
   loadingScreenRegisteredRef: React.MutableRefObject<boolean>;
 }
 
-export type SandpackStatus = "initial" | "idle" | "running" | "timeout";
+export type SandpackStatus =
```
danilowoz, line 60: Have you faced any issue with this type?

danilowoz (commented):
Hey @JayaKrishnaNamburu! 
In general, it looks good to me, I just pointed out a few UX issues and some questions.

JayaKrishnaNamburu:
Sure, will solve those this weekend 😄

danilowoz:
@JayaKrishnaNamburu forgot to mention, make sure to introduce this new feature in the documentation ✌️

sandpack-react/src/contexts/sandpackContext.tsx
JayaKrishnaNamburu: Not exactly using inside the `sandpack`. But more like to expose this functionality from outside. At the moment it's not possible to remove a file from the bundler.

sandpack-react/src/components/FileTabs/index.tsx
JayaKrishnaNamburu: I mean just to keep it uniform i let the close-icon. But actually even if clicked, it will not close the tab. I will remove the icon too 👍

sandpack-react/src/types.ts
JayaKrishnaNamburu: Yup, `done` is actually fired at the end after success in bundler. But wasn't present in the interface, so added it 👍

sandpack-react/src/components/CodeEditor/index.tsx
JayaKrishnaNamburu: Yup, reverting. In the first design i made `clossable` as default. It was needed at the time. Good catch 😄

JayaKrishnaNamburu:
@danilowoz made all the changes, please let me know if needed anything else too :) Sorry for the delay, was caught up with some stuff  😅

danilowoz:
Thank you so much @JayaKrishnaNamburu for your contribution! The changes look great! 👍

PR #76, opened 2021-07-15
feat(react): syntax inline error - CSB-735

danilowoz:
https://user-images.githubusercontent.com/4838076/125803444-744736d3-7f94-4c8a-b164-c6aed1f108ae.mov

It introduces a new prop `showInlineErrors` which basically consumes the error logs from the bundler and highlights it in the Codemirror using the class name`cm-errorLine`, which will allow customization on top of this feature.

How it works:
- Basically, a new plugin was created to handle the decoration in the editor, and this plugin listen to changes from the editor;
- Then, a new sandpack listener was introduced to listen to the bundler error and pass these messages to the editor through `annotations` changes;
- Once all errors are solved, the listener triggers a new code change (with the same input and same selection values), just to clean up the decoration - kind of a hack, but it works properly without overengineering anything.

In order to achieve our goal, it's still missing the bundler error customization, but I'm not how to proceed with it and even if we should.

Reference https://github.com/codemirror/view/blob/main/src/active-line.ts

### DIFF
--- a/sandpack-react/src/components/CodeEditor/CodeMirror.tsx
+++ b/sandpack-react/src/components/CodeEditor/CodeMirror.tsx
@@ -10,6 +10,7 @@ import { commentKeymap } from "@codemirror/comment";
 import { lineNumbers } from "@codemirror/gutter";
 import { history, historyKeymap } from "@codemirror/history";
 import { bracketMatching } from "@codemirror/matchbrackets";
+import type { Annotation } from "@codemirror/state";
 import { EditorState } from "@codemirror/state";
 import {
   highlightSpecialChars,
@@ -20,10 +21,12 @@ import {
 import type { KeyBinding } from "@codemirror/view";
 import * as React from "react";
 
+import { useSandpack } from "../../hooks/useSandpack";
 import { useSandpackTheme } from "../../hooks/useSandpackTheme";
 import type { EditorState as SandpackEditorState } from "../../types";
 import { getFileName } from "../../utils/stringUtils";
 
+import { highlightInlineError } from "./highlightInlineError";
 import {
   getCodeMirrorLanguage,
   getEditorTheme,
@@ -63,6 +66,7 @@ export const CodeMirror: React.FC<CodeMirrorProps> = ({
   const { theme, themeId } = useSandpackTheme();
   const [internalCode, setInternalCode] = React.useState<string>(code);
   const c = useClasser("sp");
+  const { listen } = useSandpack();
 
   React.useEffect(() => {
     if (!wrapper.current) {
@@ -104,6 +108,7 @@ export const CodeMirror: React.FC<CodeMirrorProps> = ({
       bracketMatching(),
       closeBrackets(),
       highlightActiveLine(),
+      highlightInlineError(),
 
       keymap.of([
         ...closeBracketsKeymap,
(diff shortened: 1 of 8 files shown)


sandpack-react/src/components/CodeEditor/CodeMirror.tsx
```diff
@@ -185,6 +195,54 @@ export const CodeMirror: React.FC<CodeMirrorProps> = ({
...
+              // Pass message to clean up inline error
+              annotations: [
+                ({
+                  type: "clean-error",
+                  value: null,
+                } as unknown) as Annotation<unknown>,
+              ],
+
+              // Trigger a doc change to remove inline error
+              changes: {
```
alexnm, line 215: question: isn't the annotation dispatch enough to trigger the change in the editor? seems like an extra operation here. don't see a big problem with it as it is not that often, but still, wondering if it couldn't introduce rendering problems in the future

sandpack-react/src/components/CodeEditor/CodeMirror.tsx
```diff
@@ -185,6 +195,54 @@ export const CodeMirror: React.FC<CodeMirrorProps> = ({
     }
   }, [code]);
 
+  React.useEffect(
+    function messageToInlineError() {
+      if (showInlineErrors) {
```
alexnm, line 200: nitpick: I'd revert the condition here to get rid of the extra indentation level

sandpack-react/src/components/CodeEditor/highlightInlineError.ts
```diff
@@ -0,0 +1,71 @@
...
+      view: EditorView,
+      message: { type: "clean-error" } | { type: "error"; value: number } | null
+    ) {
+      if (message === null || message.type === "clean-error") {
+        return Decoration.none;
+      }
+
+      if (message.type === "error") {
+        const doc = (view.state.doc as unknown) as { text: string[] };
+        const position = doc.text.reduce((acc, curr, i) => {
```
alexnm, line 55: I assume this computes the position based on the line number, can you leave some comments around the logic to make sure we don't introduce bugs later one when changing things?

alexnm (approved):
Awesome work, left some comments but this looks really good!

sandpack-react/src/components/CodeEditor/CodeMirror.tsx
danilowoz: Unfortunately, the annotation isn't enough to clean up the old decorations, that's why it needs to trigger a new change to redraw all lines. This is the biggest downside of this implementation, but I wasn't able to figure out another way to do it.

sandpack-react/src/components/CodeEditor/utils.ts
```diff
@@ -27,6 +27,10 @@ export const getEditorTheme = (theme: SandpackTheme): Extension =>
       backgroundColor: hexToCSSRGBa(theme.palette.activeBackground, 0.5),
     },
 
+    ".cm-errorLine": {
+      backgroundColor: hexToCSSRGBa(theme.palette.errorBackground, 0.2),
```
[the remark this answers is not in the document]
DannyRuchtie: looks good!
We probably want to align Sandpack one day with the new redesign
Ze is still working on the light theme for v2, but I would not wait for it now.

PR #81, opened 2021-08-13
Feat/new syntax highlight and decorators - CSB-859

danilowoz:
## Syntax-highlight
The original CodeViewer implementation used to use the Prism package as a syntax-highlight component, but its tokenization has many differences between the Codemirror leading to subtle differences between the syntax-highlight of these components. This new implementation replaces Prism to the Codemirror itself in a read-only mode, which ensures that the syntax-highlight will always match, both on CodeEditor and CodeViewer.

## New: decorators

This PR also introduces the ability to create decorators in the **CodeViewer**, and provides a way to set a `className` of portions of code or the whole line. 

**API**
```tsx
<SandpackCodeViewer
  showLineNumbers
  decorators={[
    { className: "highlight", line: 1 },
    { className: "highlight", line: 9 },
    { className: "widget", line: 12, startColumn: 26, endColumn: 38 },
  ]}
/>
```

**Preview**
Screenshot 2021-08-13 at 10 37 25


sandpack-react/package-lock.json
```diff
@@ -0,0 +1,904 @@
+{
```
alexnm, line 1: I think this should be removed in favor of yarn.lock?

alexnm (approved):

Issue #80, opened 2021-08-09
webpack5 import error: Should not import the named export 'version'.'replace' (imported as 'version') from default-exporting module (only default export is available soon)


zzg-forward:
webpack5 import error: Should not import the named export 'version'.'replace' (imported as 'version') from default-exporting module (only default export is available soon)

zzg-forward:
fix:
import packageInfo from "../package.json";
const version = packageInfo.version;

alexnm:
Hey, thanks for the suggestion, we're trying to fix this for the next release of sandpack.

danilowoz:
Hey, we believe that this issue has been solved in the latest release (v0.1.13); please let me know if this bug still persists.

Thanks for reporting it!

Issue #66, opened 2021-06-10
ReferenceError: process is not defined (Docusaurus)


pomber:
Hi. I was testing sandpack-react with Docusaurus, but there's a fatal error as soon as I import a sandpack component:

```error
client.js?e6b8:10 Uncaught (in promise) ReferenceError: process is not defined
    at eval (webpack-internal:///./node_modules/@codesandbox/sandpack-client/esm/client.js:21)
    at Module../node_modules/@codesandbox/sandpack-client/esm/client.js (:3000/vendors-node_modules_codesandbox_sandpack-react_dist_esm_presets_SandpackRunner_js.js:23)
    at __webpack_require__ (:3000/runtime~main.js:36)
    at fn (:3000/runtime~main.js:339)
    at eval (webpack-internal:///./node_modules/@codesandbox/sandpack-react/dist/esm/contexts/sandpackContext.js:8)
    at Module../node_modules/@codesandbox/sandpack-react/dist/esm/contexts/sandpackContext.js (:3000/vendors-node_modules_codesandbox_sandpack-react_dist_esm_presets_SandpackRunner_js.js:143)
    at __webpack_require__ (:3000/runtime~main.js:36)
    at fn (:3000/runtime~main.js:339)
    at eval (webpack-internal:///./node_modules/@codesandbox/sandpack-react/dist/esm/hooks/useSandpack.js:6)
    at Module../node_modules/@codesandbox/sandpack-react/dist/esm/hooks/useSandpack.js (:3000/vendors-node_modules_codesandbox_sandpack-react_dist_esm_presets_SandpackRunner_js.js:183)
    at __webpack_require__ (:3000/runtime~main.js:36)
    at fn (:3000/runtime~main.js:339)
    at eval (webpack-internal:///./node_modules/@codesandbox/sandpack-react/dist/esm/common/Layout.js:8)
    at Module../node_modules/@codesandbox/sandpack-react/dist/esm/common/Layout.js (:3000/vendors-node_modules_codesandbox_sandpack-react_dist_esm_presets_SandpackRunner_js.js:79)
    at __webpack_require__ (:3000/runtime~main.js:36)
    at fn (:3000/runtime~main.js:339)
    at eval (webpack-internal:///./node_modules/@codesandbox/sandpack-react/dist/esm/presets/SandpackRunner.js:7)
[...]

PR #85, opened 2021-08-19
Esbuild - CSB-526

danilowoz:
This replaces how we build the packages, from `tsc` to `esbuild`, which must have pretty much the same output but give us the ability to replace the environment variable. Plus, it is way faster than the previous implementation! 

It potentially solves #66 and #80


sandpack-client/build.js
```diff
@@ -0,0 +1,37 @@
...
+    ...(package.peerDependencies || {}),
+  }),
+};
+
+build({
+  ...options,
+  format: "esm",
+  outdir: "dist/esm",
+  target: "es2019",
+}).catch(() => process.exit(1));
```
DeMoorJasper, line 30: Should this catch also log the error of does esbuild send this to stderr before throwing anyway?

DeMoorJasper (approved):

sandpack-client/build.js
danilowoz: `tsc` was throwing the errors anyway, but it's a good idea to log the `esbuild` error as well, as it contains more details.

PR #86, opened 2021-08-20
ci: add workflows and commitlint

danilowoz:


sandpack-client/gulpfile.js
```diff
@@ -1,14 +1,21 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+const del = require("del");
 const gulp = require("gulp");
 const removeSourcemaps = require("gulp-remove-sourcemaps");
 
-gulp.task("copy-sandbox", () =>
+const dist = "./sandpack/";
+const paths = process.env.CI
```
alexnm, line 7: I wanted to suggest here we drop gulp and find a different way of coping over the bundler, but on the principle: "If it ain't broke, don't fix it", let's leave it like that for now

package.json
```diff
@@ -4,14 +4,14 @@
...
-    "sandpack-react",
-    "sandpack-playground"
+    "sandpack-react"
   ],
   "description": "",
   "scripts": {
     "lint": "eslint '**/*.ts?(x)' --fix",
     "format": "prettier --write '**/*.ts?(x)'",
     "build": "yarn workspaces run build",
+    "build:publish": "yarn workspaces run build:publish",
```
alexnm, line 14: can you document how the publish process works in the Readme file? I know we talked about the semantic release a while ago, but it make sense for the flow to be described for everyone that wants to contribute on the long term

lerna.json
```diff
@@ -0,0 +1,19 @@
+{
+  "packages": [
+    "sandpack-client",
+    "sandpack-react"
+  ],
+  "npmClient": "yarn",
+  "version": "independent",
```
alexnm, line 7: Not sure if this is an intentional change, but I second the decision to not have the same version for the react and the client libraries

.github/workflows/release.yml
```diff
@@ -0,0 +1,88 @@
...
+      - name: Setup | Checkout
+        uses: actions/checkout@v2
+        with:
+          # pulls all commits (needed for lerna / semantic release to correctly version)
+          fetch-depth: "0"
+
+      - name: Setup | Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: "12.22.4"
```
alexnm, line 62: are we locked with sandpack on node v12?

alexnm (approved):
looks good, left some comments

.github/workflows/release.yml
danilowoz: Yes, we do. I think `yarn` requires at least 12, but I picked this values from here https://github.com/codesandbox/sandpack/blob/main/.nvmrc

package.json
```diff
@@ -4,14 +4,14 @@
   "private": true,
   "workspaces": [
     "sandpack-client",
-    "sandpack-react",
-    "sandpack-playground"
```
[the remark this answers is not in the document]
danilowoz: Yes, we did. Actually, Jaspes is no longer using this package and it's completely outdated.

PR #87, opened 2021-08-23
fix(react): update clients after reset files

danilowoz:


sandpack-react/src/contexts/sandpackContext.tsx
```diff
@@ -521,15 +521,18 @@ class SandpackProvider extends React.PureComponent<
...
+      }),
+      this.updateClients
+    );
   };
 
   resetAllFiles = (): void => {
     const { files } = getSandpackStateFromProps(this.props);
 
-    this.setState({ files });
+    this.setState({ files }, this.updateClients);
```
alexnm, line 535: at a certain point, this calls for a refactor to get rid of the class implementation and rely on hooks to handle state transitions and effects, but for now it works well, let's focus on delivering

alexnm (approved):

PR #90, opened 2021-08-25
Imp/GitHub theme

alexnm:
* lineHeight exposed on theme
* syntax highlight updated to differentiate between function calls and property access.

### DIFF
--- a/sandpack-react/src/components/CodeEditor/utils.ts
+++ b/sandpack-react/src/components/CodeEditor/utils.ts
@@ -1,4 +1,4 @@
-import { HighlightStyle, tags } from "@codemirror/highlight";
+import { HighlightStyle, tags, Tag } from "@codemirror/highlight";
 import { css } from "@codemirror/lang-css";
 import { html } from "@codemirror/lang-html";
 import { javascript } from "@codemirror/lang-javascript";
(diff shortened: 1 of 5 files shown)


sandpack-react/src/components/CodeEditor/utils.ts
```diff
@@ -90,22 +91,29 @@ export const getSyntaxHighlight = (theme: SandpackTheme): HighlightStyle =>
...
     },
     {
       tag: [tags.literal, tags.inserted],
       ...getSyntaxStyle(theme.syntax.string ?? theme.syntax.static),
     },
-    {
-      tag: tags.propertyName,
-      ...getSyntaxStyle(theme.syntax.property),
-    },
+    // {
```
danilowoz, line 112: Should it be here?

sandpack-react/src/components/CodeEditor/utils.ts
```diff
@@ -90,22 +91,29 @@ export const getSyntaxHighlight = (theme: SandpackTheme): HighlightStyle =>
...
     },
     {
-      tag: tags.typeName,
+      tag: tags.tagName,
       ...getSyntaxStyle(theme.syntax.tag),
     },
     { tag: tags.variableName, ...getSyntaxStyle(theme.syntax.plain) },
     {
+      // Highlight function call
+      tag: tags.function(tags.variableName),
```
danilowoz, line 100: Great!

sandpack-react/src/components/CodeEditor/utils.ts
```diff
@@ -90,22 +91,29 @@ export const getSyntaxHighlight = (theme: SandpackTheme): HighlightStyle =>
       ...getSyntaxStyle(theme.syntax.static),
     },
     {
-      tag: tags.typeName,
+      tag: tags.tagName,
```
danilowoz, line 94: Is this right? Shouldn't we keep the same?

alexnm: looks to have the same effect, tagName is more suggestive

PR #92, opened 2021-09-16
chore: remove tsignore with update tsconfig

uyarn:


danilowoz:
Hey @uyarn thanks for your contribution, but in order to solve this problem we came with another solution in this PR https://github.com/codesandbox/sandpack/pull/85, which aims to improve the builder of the project.

uyarn:
ok cool

PR #104, opened 2021-10-06
fix(codeviewer): refactor ready-only mode

danilowoz:


sandpack-react/src/styles/index.css
```diff
@@ -117,6 +117,7 @@
 
 .sp-cm {
   padding: var(--sp-space-4) 0;
+  margin: 0;
```
alexnm, line 120: why is this added here?

alexnm (approved):

PR #110, opened 2021-10-14
Codeviewer: mobile issues

danilowoz:


sandpack-react/src/styles/index.css
```diff
@@ -137,8 +137,8 @@
   padding-right: var(--sp-space-2);
 }
 
-.sp-cm.sp-pristine .cm-scroller {
-  overflow: hidden;
```
danilowoz, line 141: @alexnm do you recall the reason why this `overflow`? Actually, this is causing some scroll issues on the desktop as well, turns out it is very hard to focus on the div and scroll it. Once I removed it, both desktop and mobile were working as expected, but I can't see what is side-effects of this change.

alexnm: Yes, I remember. this was added to keep the editor from hijacking the scroll as you scroll into the page.

alexnm: ok, then it's safe to remove this

PR #114, opened 2021-10-30
feat(template): add react typescript

### LINKED ISSUE #108: Add react-ts template
Would be useful to have a predefined `template` based on a `create-react-app` with out-of-the-box support for typescript

andreifloricel:

Resolves #108 

## What is the new behavior?

Provide a predefined template for react with out-of-the-box support for typescript

Extended Storybook

## Checklist

- [x] Documentation "N/A"
- [x] Ready to be merged

The public API of 'sandpack-client' has changed (added new param for 'devDependencies'). Not sure how the release process works but the `ts-ignore` from `sandpackUtils.ts` has to go.
Also prettier added some formatting but I can only assume the previous code was incorrectly formatted.

### DIFF
--- a/sandpack-react/src/templates/react-typescript.ts
+++ b/sandpack-react/src/templates/react-typescript.ts
@@ -1,0 +1,86 @@
+import type { SandboxTemplate } from "../types";
+
+export const REACT_TYPESCRIPT_TEMPLATE: SandboxTemplate = {
+  files: {
+    "tsconfig.json": {
+      code: `{
+    "include": [
+        "./**/*"
+    ],
+    "compilerOptions": {
+        "strict": true,
+        "esModuleInterop": true,
+        "lib": [
+            "dom",
+            "es2015"
+        ],
+        "jsx": "react-jsx"
+    }
+}`,
+    },
+    "/App.tsx": {
+      code: `export default function App() {
+  return <h1>Hello World</h1>
+}
+`,
+    },
+    "/index.tsx": {
+      code: `import React, { StrictMode } from "react";
+import ReactDOM from "react-dom";
+import "./styles.css";
+
+import App from "./App";
+
+const rootElement = document.getElementById("root");
+ReactDOM.render(
+  <StrictMode>
+    <App />
+  </StrictMode>,
+  rootElement
+);`,
+    },
+    "/styles.css": {
+      code: `body {
+  font-family: sans-serif;
+  -webkit-font-smoothing: auto;
+  -moz-font-smoothing: auto;
+  -moz-osx-font-smoothing: grayscale;
+  font-smoothing: auto;
+  text-rendering: optimizeLegibility;
+  font-smooth: always;
+  -webkit-tap-highlight-color: transparent;
+  -webkit-touch-callout: none;
+}
+
+h1 {
+  font-size: 1.5rem;
+}`,
+    },
+    "/public/index.html": {
+      code: `<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Document</title>
+  </head>
+  <body>
+    <div id="root"></div>
+  </body>
+</html>`,
+    },
+  },
+  dependencies: {
+    react: "^17.0.0",
+    "react-dom": "^17.0.0",
+    "react-scripts": "^4.0.0",
+  },
+  devDependencies: {
+    "@types/react": "^17.0.0",
+    "@types/react-dom": "^17.0.0",
+    typescript: "^4.0.0",
+  },
+  entry: "/index.tsx",
+  main: "/App.tsx",
+  environment: "create-react-app",
+};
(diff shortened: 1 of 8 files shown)


danilowoz:
This really is awesome! I'll review and leave my thoughts as soon as I can. @alexnm who is the best person to review the `client` changes?

andreifloricel:
Thanks! I just pushed a small update to the client changes as I noticed that my previous implementation was a bit over-enginereed.

sandpack-react/src/utils/sandpackUtils.ts
```diff
@@ -78,6 +78,8 @@ export const getSandpackStateFromProps = (
   const files = addPackageJSONIfNeeded(
     projectSetup.files,
     projectSetup.dependencies || {},
+    projectSetup.devDependencies || {},
+    // @ts-ignore FIXME: remove this as soon as the sandpack-client typings are up-to-date
```
alexnm, line 82: if you are running the project locally, these packages should be linked together as part of the workspace and you should get the current typing for the code in `sandpack-client`

sandpack-react/src/templates/react-typescript.ts
```diff
@@ -0,0 +1,86 @@
...
+  </body>
+</html>`,
+    },
+  },
+  dependencies: {
+    react: "^17.0.0",
+    "react-dom": "^17.0.0",
+    "react-scripts": "^4.0.0",
+  },
+  devDependencies: {
```
alexnm, line 78: not 100% sure, but I think create-react-app works with typescript as long as you provide a tsconfig file. @CompuIves I'm curious what's your take on differentiating `dependencies` and `devDependencies` in the sandpack API. Are there any differences in how they are treated in the bundler?

alexnm (commented):
Great work Andrei! Thank you for your contribution, left a few notes, then we can soon merge this.

sandpack-react/src/templates/react-typescript.ts
```diff
@@ -0,0 +1,86 @@
...
+        "lib": [
+            "dom",
+            "es2015"
+        ],
+        "jsx": "react-jsx"
+    }
+}`,
+    },
+    "/App.tsx": {
+      code: `export default function App() {
```
danilowoz, line 22: Could you type this component? It would be important for the user to understand they're using a different template.

sandpack-react/src/templates/react-typescript.ts
andreifloricel: I tried to mimic the "React Typescript" template.
IMO this is not only about the React types but also about types for 3rd party libraries which might be included in the sandbox.
If custom `devDependencies` are not yet supported, then maybe a possible workaround for sandpack would be to provide the `@types` as runtime `dependencies`?

andreifloricel: Hi @alexnm, are there any news about this topic?

CompuIves: Heya! Yeah there is a difference, we generally don't install `devDependencies`, because most tools in there were build tools, while sandpack is a build tool, and it increased bandwidth usage quite a bit.

However, there were plans to support `devDependencies` by going through the application code and detecting which dependencies are used, and only using those. But that would've required us to rewrite the parsing logic (only install dependencies after first parse). So that's still on the backlog.

CompuIves: But! The general idea is that `devDependencies` should be there, because when you download the repository locally you _need_ to be able to run it as well. So it's good that it's added here.

CompuIves: And, `@types/...` is used in CodeSandbox editor to dynamically download type definitions, even if it's defined in `devDependencies`. This is not used in Sandpack though, since it only cares about bundling. I would keep it here!

CompuIves (approved):
Super nice addition! Happy to see React-TS in here 😄 .

andreifloricel:
Thank you! Glad I can contribute something, after using codesandbox so many times :)
I'm not sure how the PR process goes forward. I see that the Vercel deployment failed, but I don't have access to its logs

andreifloricel:
Hmm, as I created #127 and checked the existing codesandbox & sandpack templates I realised that all sandpack templates (**except react**) follow the sandbox structure in the sense that source files are placed under the `/src` directory.:
ex. Vue https://codesandbox.io/s/ecstatic-ptolemy-gxt88 & https://github.com/codesandbox/sandpack/blob/main/sandpack-react/src/templates/vue.ts#L5

Maybe I should adjust the `react-ts` template to have the same structure and be consistent with all the other ones? 
@CompuIves @alexnm what's your take on this?

alexnm:
> Hmm, as I created #127 and checked the existing codesandbox & sandpack templates I realised that all sandpack templates (**except react**) follow the sandbox structure in the sense that source files are placed under the `/src` directory.: ex. Vue https://codesandbox.io/s/ecstatic-ptolemy-gxt88 & https://github.com/codesandbox/sandpack/blob/main/sandpack-react/src/templates/vue.ts#L5
> 
> Maybe I should adjust the `react-ts` template to have the same structure and be consistent with all the other ones? @CompuIves @alexnm what's your take on this?

I don't know if we have any constraints here, but I would follow the template that is used by codesandbox.io for now

alexnm (approved):

andreifloricel:
Agreed, I adjusted it.

danilowoz (approved):

PR #118, opened 2021-11-02
chore: bootstrap project with create-next-app

olarclara:
Creates landing page project with `create-next-app + ts` inside `/website/landing`.


danilowoz (approved):
Sounds good! Please, create another branch to merge this PR while the website is a work in progress.

---

**Open questions**
1. Do you have anything in mind for styles? Take a look at how Stitches works in another project, it might be a great fit here as well;
2. Can we consume the design tokens from our design system? By the first step, we can copy and paste, but it would be cool to consume it as a package;
3. I'm very excited to see how `next lint` works in the latest Nextjs version. Can you set up some actions to lint the code?
4. Can we use MDx or something else to make the context easier to update? This is not a hard goal, but I'd encourage avoiding hard-coded content in the components.

olarclara:
@danilowoz sure, created a feature branch `website/landing`.

1. I think Stitches is a great option.
2. Sounds good to me, is the design system published already?
3. Do you mean on top of this check: https://github.com/codesandbox/sandpack/blob/main/.github/workflows/checks.yml#L6?
4. Sure, I'll research a good approach to avoid hard-coded content.

PR #120, opened 2021-11-03
fix(open in codesandbox button): optimize for long sandoxes

danilowoz:
This introduces a breaking change due to `useCodeSandboxLink` is no longer used to generate the URL. From now on, the `UnstyledOpenInCodeSandboxButton` is responsible for managing the URL or the form that will create a new sandbox in the CodeSandbox.

In the previous approach, all the parameters to create a sandbox in the CodeSandbox were attached directly in the URL, making it easily reach the `request too long (401)`. That's why we deprecated the hook in favor of using a whole new component, which is responsible for either using an anchor (for short sandboxes) or a form-POST (for long sandboxes). 

The `UnstyledOpenInCodeSandboxButton` is a basic component that doesn't carry any style, so if you want a ready-to-use component, use the `OpenInCodeSandboxButton` instead, which contains the CodeSandbox logo and the same functionality. 

---

Migration guide:
```diff
import {
--  useCodeSandboxLink,
++  UnstyledOpenInCodeSandboxButton
}
const CustomOpenInCSB = () => {
--  const url = useCodeSandboxLink();
--  return <a href={url}Open in CodeSandbox</a>
++  return <UnstyledOpenInCodeSandboxButton>pen in CodeSandbox</UnstyledOpenInCodeSandboxButton>
   );
};
```

CSB-519


sandpack-react/src/common/BareCodeSandboxButton.tsx
```diff
@@ -0,0 +1,96 @@
...
+  // Register the usage of the codesandbox link
+  React.useEffect(function registerUsage() {
+    sandpack.openInCSBRegisteredRef.current = true;
+  }, []);
+
+  /**
+   * This is a safe limit to avoid too long requests (401),
+   * as all parameters are attached in the URL
+   */
+  if (paramsValues.length > 1500) {
```
danilowoz, line 70: What do you think about this conditional? Doesn't it sound arbitrary?

sandpack-react/src/index.ts
```diff
@@ -12,10 +12,10 @@ export * from "./common/Layout";
 export * from "./common/Stack";
 export * from "./common/ErrorOverlay";
 export * from "./common/LoadingOverlay";
+export * from "./common/BareCodeSandboxButton";
```
alexnm, line 15: I would call this `UnstyledOpenInCodeSandboxButton`, to first keep consistency with the `OpenInCodeSandboxButton` name and to better suggest what it is

sandpack-react/src/common/BareCodeSandboxButton.tsx
alexnm: limit is 2000 for the entire URL as far as I know, I think it's a fair value, as there are several other elements of the URL maybe extract it in a constant somewhere? with an explanation (brief) about the value

sandpack-react/src/common/BareCodeSandboxButton.tsx
```diff
@@ -0,0 +1,96 @@
...
+export const BareCodeSandboxButton: React.FC<{ className?: string }> = ({
+  children,
+  ...props
+}) => {
+  const [paramsValues, setParamsValues] = React.useState("");
+  const { sandpack } = useSandpack();
+  const formRef = React.useRef<HTMLFormElement>(null);
+
+  React.useEffect(
+    function debounce() {
```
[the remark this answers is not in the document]
alexnm: was this implemented before? I don't remember now. Is it precaution to not call this too often?

alexnm (approved):
Left a few comments, but nothing to block the PR if you have other priorities right now

daryl-cecile:
It seems the export for `useCodeSandboxLink` in '_sandpack-react/src/hooks/index.ts_' needs to be removed

andreifloricel:
Hi @danilowoz, I just stumbled on this limitation (401 on long requests) and wanted to open an issue but than noticed this PR :)
Is this fix planned for a coming release?

website/docs/docs/advanced-usage/components.md
```diff
@@ -173,6 +173,31 @@ For situations when you strictly want to show some code and run it in the browse
   </SandpackLayout>
 </SandpackProvider>
 
+## UnstyledOpenInCodeSandboxButton & OpenInCodeSandboxButton
+
+You can build a custom button that opens the sandpack files in a new tab on https://codesandbox.io. Let's the use `UnstyledOpenInCodeSandboxButton` for that:
```
sannek, line 178:
```suggestion
You can build a custom button that creates a new sandbox from the sandpack files. It will include any edits made in the Sandpack editor, so it is a great way to persist your changes. The created sandbox will open on [CodeSandbox](https://codesandbox.io) in a new tab.

Let's use the `UnstyledOpenInCodeSandboxButton` as an example:
```

sannek (commented):
Some small docs suggestion ❤️  Looks great tho!

PR #122, opened 2021-11-03
feat: design tokens

olarclara:

Adds tokens copied from the existing design system. Next PR will include a ThemeProvider to inject global styles such as fonts and resets.


website/landing/styles/breakpoints.ts
```diff
@@ -0,0 +1,15 @@
+export type Breakpoint = "sm" | "md" | "lg" | "xg";
```
danilowoz, line 1: @smhutch what if we publish the design tokens to keep all applications up to date with the latest changes?

danilowoz (approved):

PR #141, opened 2021-11-12
docs: api reference

danilowoz:


sandpack-react/src/contexts/sandpackContext.tsx
```diff
@@ -130,11 +132,17 @@ class SandpackProvider extends React.PureComponent<
...
+      React.createRef<boolean>() as React.MutableRefObject<boolean>;
+    this.openInCSBRegistered =
+      React.createRef<boolean>() as React.MutableRefObject<boolean>;
+    this.loadingScreenRegistered =
+      React.createRef<boolean>() as React.MutableRefObject<boolean>;
   }
 
+  /**
+   * @hidden
+   */
```
alexnm, line 145: isn't there a better way to ignore all members inside the context here?

alexnm (approved):

PR #154, opened 2021-11-17
chore: landing ui tweaks

danilowoz:


website/landing/components/common/Clipboard.tsx
```diff
@@ -100,7 +100,7 @@ export const Clipboard: React.FC = () => {
           transition: "color .2s ease", // TODO: verify transition
           willChange: "color",
 
-          "&:hover, &:hover path": {
+          "&:hover": {
```
olarclara, line 103: Will the clipboard icon change its color without this?

danilowoz: Yes, removing the color from the next element makes the SVG inherit color from here

olarclara: Great! I asked because it didn't work when I was implementing it. :/

PR #162, opened 2021-11-19
feat(Docs) Update styles

zehfernandes:
**Old:**
 
image

**New:**

image


website/docs/docusaurus.config.js
```diff
@@ -162,14 +164,14 @@ module.exports = {
...
         },
-        items: [
-          { to: "/releases", label: "Releases", position: "right" },
-          {
-            href: "https://github.com/codesandbox/sandpack",
-            label: "GitHub",
-            position: "right",
-          },
-        ],
+        // items: [
```
danilowoz, line 167: you won't go to heaven behaving like this

danilowoz (approved):
Oh you made it! Thanks so much
So I believe we can close most of these issues https://linear.app/codesandbox/issue/CSB-1665/sandpack-docs-layout-and-configuration

website/docs/docusaurus.config.js
zehfernandes: spotted! 😅

PR #177, opened 2021-11-24
Vertical component for nested sandpack on the documentation

zehfernandes:
@sannek here is the vertical component. I would like to customize more the second preview but that will require not used the Sandpack component and loose all the default styles. Maybe in a next iteration?

image


danilowoz (approved):
@zehfernandes is it ready to be merged? It looks good to me

sannek:
Awesome, thanks Zeh!
I don't think - at least for now - we should further customize the preview. It might look nicer, but will make it no longer a good preview of what it would _really_ look like when people use it in their projects.

PR #130, opened 2021-11-10
feat(template): add vanilla typescript

andreifloricel:
Resolves #127 

## What is the new behavior?

Provides a sandpack template with out-of-the-box support for typescript.
Technically it is similar with the existing "vanilla" sandpack template, but with `.ts` files instead of `.js`.
The resulted sandbox follows the "Vanilla Typescript" codesandbox template.

## What steps did you take to test this? This is required before we can merge, make sure to test the flow you've updated.

Extended Storybook

## Checklist

- [ ] Documentation "N/A"
- [x] Ready to be merged

PR #182, opened 2021-11-25
feat(template): svelte 

paolotiu:

It provides better syntax highlighting for `.svelte` files

Currently, `.svelte` files are being associated as javascript files.

## What is the new behavior?

Instead of javascript, `.svelte` files will be associated with html files since they are more similar.

1. Added a case for svelte in the extensions switch case.

## Checklist

- [ ] Documentation N/A
- [x] Ready to be merged

Loving the project!! <3

### DIFF
--- a/sandpack-react/src/components/CodeEditor/utils.ts
+++ b/sandpack-react/src/components/CodeEditor/utils.ts
@@ -140,6 +140,7 @@ export const getCodeMirrorLanguage = (
       return javascript({ jsx: true, typescript: true });
     case "vue":
     case "html":
+    case "svelte":
       return html();
     case "css":
     case "scss":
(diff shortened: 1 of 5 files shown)


danilowoz:
Hey @paolotiu thanks for your contribution! 
But what if we go further and implement a Svelte template? If so, you could use #130 as a recipe to introduce it on Sandpack.

Hope we can make it work!

paolotiu:
> Hey @paolotiu thanks for your contribution! But what if we go further and implement a Svelte template? If so, you could use #130 as a recipe to introduce it on Sandpack.
> 
> Hope we can make it work!

Sure!

paolotiu:
@danilowoz the vanilla svelte template was straightforward, but I think it would be great if there would also be a typescript version. I tried for a few hours, but the only working sandbox I created used containers which I _think_ isn't supported right now. How would I go about implementing this?

danilowoz:
Thanks, @paolotiu, for your contribution! 
Unfortunately, Sandpack still doesn't support Typescript, and there are a few limitations around this issue. But hopefully, we will make it work soon!

paolotiu:
> Thanks, @paolotiu, for your contribution! Unfortunately, Sandpack still doesn't support Typescript, and there are a few limitations around this issue. But hopefully, we will make it work soon!

Got it! Nevertheless, this is an awesome project! Loving it so far

PR #183, opened 2021-11-25
landing: responsive fixes

danilowoz:


website/landing/components/Users/Users.tsx
```diff
@@ -39,6 +40,7 @@ const UserLink = styled("a", {
 
 export const Users: React.FC = () => {
   const content = config.users;
+  const shouldAnimate = useBreakpoint("bp2");
```
olarclara, line 43: The main reason for the animation was to avoid the hard transition next/image does. Can we keep the behavior? And then, use the `shouldAnimate` as a flag for the intersection observer:

```
  const { ref: listRef, inView } = useInView({
    threshold: [0, 1],
    triggerOnce: !shouldAnimate
  });
```

This way it won't animate back and forth when scrolling but it will keep the fade for the first load.

PR #231, opened 2021-12-03
fix(react): ensure hidden files even for templates

### LINKED ISSUE #222: Bug: When only one file is listed, "hidden" doesn't work
# Bug report

## Packages affected

- [] sandpack-client
- [x] sandpack-react

## Description of the problem

The "hidden" property doesn't seem to work when only passing in one file.

```
    <Sandpack
        theme="dark"
        template="react"
        files={{
            '/hiddenFile.js': {
                code: ``,
                hidden: true,
            },
        }}
    />
```

Screen Shot 2021-12-02 at 9 19 48 PM

If I add another non-hidden file to the file-list, the hidden file will disappear.

## What were you doing when the problem occurred?

Use the code snippet above, with:
```
import { Sandpack } from '@codesandbox/sandpack-react'
import '@codesandbox/sandpack-react/dist/index.css'
```

### Your Environment

| Sandpack-client version |              |
| Sandpack-react version  | ^0.9.8 |
| Browser                 | Chrome |
| Operating System        |Mac |

danilowoz:
Ensure that `hidden` files don't appear when `openPaths` is omitted.
Fix #222


olarclara (approved):

sandpack-react/src/utils/sandpackUtils.ts
```diff
@@ -56,7 +56,13 @@ export const getSandpackStateFromProps = (
 
   if (openPaths.length === 0) {
     // If no files are received, use the project setup / template
-    openPaths = Object.keys(projectSetup.files);
+    openPaths = Object.keys(projectSetup.files).reduce<string[]>((acc, key) => {
+      if (projectSetup.files[key].hidden) {
+        return acc;
+      }
+
+      return [...acc, key];
```
lbogdan, line 64: This creates a new array on every iteration, I'd suggest:

```js
acc.push(key);
return acc;
```

lbogdan: Actually, considering ^^, the code can become even simpler:

```js
    openPaths = Object.keys(projectSetup.files).reduce<string[]>((acc, key) => {
      if (!projectSetup.files[key].hidden) {
        acc.push(key);
      }

      return acc;
    }, []);
```

PR #239, opened 2021-12-03
fix(types): react components not actually accepting children

feychenie:

TS types fix

Most of the react components implemented the React.FC type, implying they accept `children` while they actually don't.

React.FC in still supported by react for backwards compatibility, but has been recognised officially as an anti-pattern for a while.

Here are some readings about this:
- https://github.com/typescript-cheatsheets/react#function-components (section "Why is `React.FC` discouraged?")
- https://kentcdodds.com/blog/how-to-write-a-react-component-in-typescript
- https://dev.to/wojciechmatuszewski/top-three-react-typescript-pitfalls-50l8
- https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/

## What is the new behavior?

Components are not falsely types as supporting children.

Additionally, react imports have been unified to the `import * from React` pattern that seem to be used in the majority of the other files.

1. Checked in the code that the affected components are not rendering props.children

## Checklist

- [ ] Documentation
- [x] Ready to be merged

### DIFF
--- a/sandpack-react/src/components/CodeEditor/CodeEditor.stories.tsx
+++ b/sandpack-react/src/components/CodeEditor/CodeEditor.stories.tsx
@@ -1,5 +1,5 @@
 import type { Story } from "@storybook/react";
-import React from "react";
+import * as React from "react";
 
 import { SandpackProvider } from "../../contexts/sandpackContext";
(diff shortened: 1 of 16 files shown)


danilowoz:
Nice catch! Thanks, @feychenie, it looks good to me 🎉

PR #241, opened 2021-12-05
fix(react-context): expose clients 

danilowoz:
It exposes the `clients` map under the same contexts, which gives the possibility to interact with all clients under the same provider. Providing access to the clients allows to communicate with the iframe elements and pass messages.

This is the safest way to interact with clients and their iframes, as Sandpack context ensures access to all Preview that has been rendered under the same provider.


sannek:
I only looked at the docs change, but that seems clear to me!

sannek (approved):

PR #269, opened 2021-12-21
docs: add faq

danilowoz:


website/docs/docs/faq.md
```diff
@@ -0,0 +1,44 @@
+---
+sidebar_position: 5
+title: FAQ
+---
+
+# Frequently asked questions
+
+**What is Sandpack?**
+
+Sandpack is an open ecosystem of components and utilities that allow you to compile and run modern frameworks in the browser. You can either use one of our predefined `components` for embedding the *CodeSandbox* experience into your projects, or you can build your own version of `sandpack`, on top of our standard components and utilities. As you walk through this guide, you will get deeper into our ecosystem.
```
sannek, line 10: > As you walk through this guide, you will get deeper into our ecosystem.

I'm not sure if this sentence make sense in the FAQ, rather than in the docs

website/docs/docs/faq.md
```diff
@@ -0,0 +1,44 @@
...
+
+**How to load local or private dependencies?**
+
+Currently, Sandpack doesn’t have a way to consume private dependencies from any kind of registry service, due to the bundler host is shared with all Sandpack consumers apps. However, you can pass local dependencies just like a regular file or using external resource API:
```
sannek, line 14:
```suggestion
Currently, Sandpack doesn’t have a way to consume private dependencies from any kind of registry service, because the bundler host is shared with all Sandpack consumers apps. However, you can pass local dependencies just like a regular file or using the external resource API:
```

website/docs/docs/faq.md
```diff
@@ -0,0 +1,44 @@
...
+      }`,
+        },
+    }}
+    template="react"
+/>
+```
+
+**Does Sandpack support NextJs or other server-side applications?**
+
+These kinds of sandboxes are backed by a container running on a server. Therefore we can't easily support this in the Sandpack library, as you'd probably have to host your own servers in that case or integrate entirely with our APIs, which likely requires logins and paying above certain limits.
```
sannek, line 36: Maybe we can change this to make it clearer that this is currently _not possible_ at all (or is it?), because this answer sounds to me like it's "possible but difficult" instead of "currently impossible", and I'm not sure what the right message is?

sannek (commented):
Very nice to add a FAQ! A few small comments, but great to have!

PR #291, opened 2022-01-04
tests(react): ssr features

danilowoz:


cypress/integration/CodeMirror.spec.js
```diff
@@ -1,7 +1,7 @@
-describe("CodeMirror", () => {
-  const languages = ["html", "javascript", "jsx", "css", "less", "vue"];
+import * as mocks from "../../sandpack-react/src/components/CodeEditor/languages-mocks";
```
danilowoz, line 1: Refactor: take the advantage of some mocks and loop through them in order to generate some snapshots.

sandpack-react/src/components/CodeEditor/CodeMirror.stories.tsx
```diff
@@ -1,174 +1,28 @@
...
-};
-</script>
-`}
-      fileType="vue"
-      id="vue"
-      initMode="immediate"
-      showLineNumbers={false}
-    />
-  </SandpackProvider>
+import { SandpackThemeProvider } from "../../contexts/themeContext";
```
danilowoz, line 5: Refactor: as same as above

PR #297, opened 2022-01-11
docs(ssr): how to use and migration guide

danilowoz:


website/docs/releases/2022-01-10-ssr.md
```diff
@@ -0,0 +1,55 @@
...
+Sandpack has introduced a new API to support server-side rendering of the CSS theme and the syntax-highlight component from version TODO, as you can read more [here](/getting-started/ssr). As a result, there are some slight changes in the current Sandpack API (more specific to the theme), where you can check as follows:
+
+## Migration guide
+
+### CSS file import
+
+Before, you had to explicitly import a CSS file, which contained the default styles for the component. From now on, all styles are dynamically injected into the head element, or you can retrieve it from the method described bellow.
+
+```diff
+ import { Sandpack } from "@codesandbox/sandpack-react";
```
sannek, line 16: This is really nice! I kept forgetting to import the CSS file!
We should make sure we have a look at the other docs, and update our `sandpack-in-sandpack` examples as well! It would suck if they get out of date already.

PR #333, opened 2022-01-26
fix(editorstate): watch changes on files instead of each method

### LINKED ISSUE #332: Bug: `editorState` doesn't track code state
# Bug report

## Packages affected

- [ ] sandpack-client
- [x] sandpack-react

## Description of the problem

The `editorState` value found on the `sandpack` instance appears to track whether or not the user has toggled between files, rather than whether they've edited the code:

https://user-images.githubusercontent.com/6692932/151000878-2a5ff0a5-de69-42ef-9662-3aa8226ee913.mp4

It's possible I've misunderstood what `editorState` is meant to track, but it seems like for most use cases, it would be much more valuable to know whether the user has tampered with the code, vs. changed tabs. 

For context: I want to show a little "*" next to the sandbox's title when the user modifies code, so that they know it's been modified. It's important because the sandboxes are in lessons, and the lesson text refers to the original code (eg. "Notice how the text is red"). If the user changes the text from red to yellow, I want to show that the code has been modified, to explain the discrepancy. The code is saved and restored, so they might've modified it in a previous session and forgotten.

### Link to sandbox: 
https://codesandbox.io/s/sandpack-editorstate-demo-mhoyb

### Your Environment

| Sandpack-client version | — |
| Sandpack-react version  | 0.13.5 |
| Browser                 | Chrome 97.0 |
| Operating System        | MacOS Monterey |

danilowoz:
Shift the approach of how the `editorState` has been updated in the Sandpack context. From now on, it watches any changes on files and then sets if the current editor has any changes performed internally. 

Closes #332

### DIFF
--- a/.codesandbox/workspace.json
+++ b/.codesandbox/workspace.json
@@ -2,11 +2,6 @@
   "default": {
     "name": "Default",
     "type": "SIDEKICK",
-    "items": [
-      { "type": "PREVIEW", "port": "3000" },
-      { "type": "PREVIEW", "port": "3001" },
-      { "type": "PREVIEW", "port": "3002" },
-      { "type": "PREVIEW", "port": "6006" }
-    ]
+    "items": [{ "type": "PREVIEW", "port": "6006" }]
   }
 }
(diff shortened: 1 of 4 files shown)


alexnm (approved):
lgtm

PR #357, opened 2022-02-04
OpenInCodeSandboxButton: preserve ?file= param for big sandboxes

sophiebits:

Ensure that the current file stays selected even if the URL is too long and we switch to a form.

The `file=` param is set only if paramsValues.length < 1500.

## What is the new behavior?

It's consistently set, so the right file is opened in the new sandbox.

[I wrote this in the GitHub web editor. Hoping to test the sandbox on this PR.]

## Checklist

- [ ] Documentation [N/A]
- [ ] Tests [I didn't see any for this feature]
- [ ] Ready to be merged


danilowoz:
Thanks a lot, @sophiebits, just did a few tweaks (prettier & eslint), but it looks good to me!

PR #377, opened 2022-02-18
fix: don't make unnecessary sync updates

gaearon:
This is a nested update. Nested updates are bad for performance. They're also always sync, which isn't good for hydration.

In this case the whole pattern is unfortunate. Instead of a nested update, this should use `getDerivedStateFromProps`. But for now, this is an OK workaround that should improve perf.


danilowoz:
Thanks, I'll address other possible improvements in this method. 

@gaearon I had the idea that `.setState` already does this conditional internally, which would avoid unnecessary updates. Am I assuming this wrong or is there any other problem there?

alameer5:
نعم انه تحديث متداخل وليس مجدي بالنفع،لذا فان النمط برمته مؤسف جدا

Issue #383, opened 2022-02-20
CodeMirror plugin crashed: Error: Ranges must be added sorted by `from` position and `startSide`


Jokcy:
# Bug report

## Packages affected

-   [ ] sandpack-client
-   [x] sandpack-react

sandpack-client not test

## Description of the problem

Decorators will fail when use inline decorators with line number small then 10.

https://codesandbox.io/s/sandpack-project-forked-vzf788?file=/data.js

See this sandbox, open the console you will see an error `CodeMirror plugin crashed: Error xxxxx`. But if you update the `line` property to 12, it will work fine.

### Link to sandbox: [link]() (optional)

https://codesandbox.io/s/sandpack-project-forked-vzf788?file=/data.js

### Your Environment

| Software                | Name/Version |
| ----------------------- | ------------ |
| Sandpack-client version |             |
| Sandpack-react version  |      latest        |
| Browser                 |       chrome       |
| Operating System        |      mac        |

danilowoz:
Hey @Jokcy, thanks for reporting this issue! 

It's a very interesting bug because CodeMirror needs an array sorted by the key `line`, so the following example should work:

```js
export default [
  { className: "highlight", line: 1 },
  {
    className: "widget",
    elementAttributes: { "data-id": "1" },
    line: 8, // Moved to between 1 and 9
    startColumn: 3,
    endColumn: 7
  },
  { className: "highlight", line: 9 }
];

```

I think we can easily fix it on the Sandpack side, would you like to give it a try?

Jokcy:
@danilowoz Yeah I would like to have a try, I will give some feedback soon.

PR #385, opened 2022-02-21
fix: sort decorators before pass to codemirror

Jokcy:

Bug fix

Pass down decorators without sort by line will cause codemirror error.

https://github.com/codesandbox/sandpack/issues/383

## What is the new behavior?

decorators without sort will work fine

I did not find component unit tests, so I edit the storybook a little to reproduce this problem, and fix the code to make it work.

## Checklist

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

By the way, is there any plan to add unit tests? I can also do some help if needed.

### DIFF
--- a/sandpack-react/src/components/CodeEditor/CodeMirror.tsx
+++ b/sandpack-react/src/components/CodeEditor/CodeMirror.tsx
@@ -147,6 +147,16 @@ export const CodeMirror = React.forwardRef<CodeMirrorRef, CodeMirrorProps>(
     const languageExtension = getLanguageFromFile(filePath, fileType);
     const langSupport = getCodeMirrorLanguage(languageExtension);
 
+    // decorators need to be sorted by `line`, otherwise it will throw error
+    // see https://github.com/codesandbox/sandpack/issues/383
+    const sortedDecorators = React.useMemo(
+      () =>
+        decorators
+          ? decorators.sort((d1, d2) => d1.line - d2.line)
+          : decorators,
+      [decorators]
+    );
+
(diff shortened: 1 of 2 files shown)


Jokcy:
Notice I changed the order of decorators in storybook to test if the fix work fine.

sandpack-react/src/components/CodeViewer/index.tsx
```diff
@@ -51,14 +51,24 @@ export const SandpackCodeViewer = React.forwardRef<
 
     const shouldShowTabs = showTabs ?? sandpack.openPaths.length > 1;
 
+    // decorators need to sort by `line`, otherwise it will throw error
+    // see https://github.com/codesandbox/sandpack/issues/383
+    const sortedDecorators = React.useMemo(
```
danilowoz, line 56: Nice! 

Could you please do this logic at the CodeMirror component level? Or maybe in the`highlighDecorators.ts`? Then we make sure that we cover all possible components that use CodeMirror.

danilowoz:
I was trying to figure out how to test it, but it's quite challenging to test CodeMirror integrations on Sandpack... I'll try to spend more time in the test suit and make this integration more testable.

sandpack-react/src/components/CodeViewer/index.tsx
Jokcy: Yeah sure, already updated, didn't drive deep into the code, so just chose the simplest way😂.

danilowoz (approved):
Thanks a lot!

PR #396, opened 2022-03-07
fix: side effect should contain css

uyarn:

- bug fix, the `@codesandbox/sandpack-react` package requires users to import the css file (`import "@codesandbox/sandpack-react/dist/index.css"`), which would be removed after tree shaking

-  style from css would be removed after tree shaking

## What is the new behavior?

-  style from css would not be removed after tree shaking

- would occur in any projects use `@codesandbox/sanpack-react` using tree shaking

## Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/sandpack-react/package.json
+++ b/sandpack-react/package.json
@@ -12,7 +12,7 @@
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
   "types": "dist/types/index.d.ts",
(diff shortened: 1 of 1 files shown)


uyarn:
have not contributed to this project for a long time, how should I do the deploy the vercel check > . <

Issue #398, opened 2022-03-07
Pages using the `Sandpack` component are generating 2 history entries on Firefox


noobinthisgame:
# Bug report

## Packages affected

-   [ ] sandpack-client
-   [x] sandpack-react

## Description of the problem

Hi ! 
while working on the ChakraUI docs and navigating through them I noticed that when I was using Firefox I had to go back twice in the history after visiting a page where we used Sanpack as Component.
This behaviour did appear on Chrome or Safari.

### What steps can we take to reproduce the problem?

1. Open Firefox
2. Go to https://chakra-ui.com/ 
3. You can see that you can go back twice in the page history
4. Go to https://chakra-ui.com/docs/components/recipes/floating-labels
5. Same behvaviour

### Link to sandbox: [link]() (optional)

Code of ChakraUI docs is open source https://github.com/chakra-ui/chakra-ui-docs

### Your Environment

| Software                | Name/Version |
| ----------------------- | ------------ |
| Sandpack-client version |      -        |
| Sandpack-react version  |      "0.13.11"      |
| Browser                 |        Firefox 95.0.2 and 97.0.1     |
| Operating System        |       Windows / MacOS       |

danilowoz:
Searching about this bug, I've found a couple of useful links:
- Possible solution: https://github.com/aFarkas/lazysizes/commit/512800de6ebddf6a5000624b67f86c4f0d77d8c6
- More details: https://stackoverflow.com/questions/2245883/browser-back-acts-on-nested-iframe-before-the-page-itself-is-there-a-way-to-av

noobinthisgame:
Thanks for solving this so quickly!! :)

PR #407, opened 2022-03-11
fix(client): prevent add route into the main page browser history

danilowoz:
**Problem:** 
The iframe assigns a new URL in a manner that adds a new entry to the browser's list of visited URLs, on Firefox. 

**Solution:** 
So, instead of manipulating the iframe dom and adding/removing the `src` attribute, it manipulates the iframe browser history.

Fixes https://github.com/codesandbox/sandpack/issues/398

Useful links:
- https://github.com/aFarkas/lazysizes/commit/512800de6ebddf6a5000624b67f86c4f0d77d8c6
- [stackoverflow.com/questions/2245883/browser-back-acts-on-nested-iframe-before-the-page-itself-is-there-a-way-to-av](https://stackoverflow.com/questions/2245883/browser-back-acts-on-nested-iframe-before-the-page-itself-is-there-a-way-to-av)

### DIFF
--- a/sandpack-client/src/client.ts
+++ b/sandpack-client/src/client.ts
@@ -157,9 +157,12 @@ export class SandpackClient {
       );
     }
 
-    this.iframe.src = options.startRoute
+    const urlSource = options.startRoute
       ? new URL(options.startRoute, this.bundlerURL).toString()
       : this.bundlerURL;
+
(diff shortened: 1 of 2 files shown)


DeMoorJasper (approved):
Looks good to me, don't think this will have any unwanted side-effects

alexnm (approved):
lgtm

Issue #423, opened 2022-03-25
Allow user to pass clientId to SandpackPreview


malerba118:
# Tell us how you think we can improve Sandpack
I would like to augment SandpackPreview in user-land to allow message passing with a specific iframe instance. This will let me, for example, have a slider in the host site that passes its current value to the iframe and will allow the iframe to communicate back. The api i'm trying to achieve is something like.

```jsx
<AugmentedSandpackPreview 
   data={{ opacity: .2, scale: 1.5 }} 
   onMessage={(e) => console.log(e.type, e.data)} 
/>
```

But in order to do this i think i need access to the clientId that's generated within SandpackPreview.

If i had control over/access to clientId then i could do something like this 

```jsx
const AugmentedSandpackPreview = ({ data, ...otherProps }) => {
  const [clientId] = useState(() => nanoid())

  const { sandpack } = useSandpack()

  const iframe = sandpack.clients[clientId].iframe

  useEffect(() => {
     iframe.contentWindow.postMessage(JSON.stringify(data), '*')
  }, [data])

  return <SandpackPreview clientId={clientId} {...otherProps} />
}
```

Alternatively access to the iframe through a ref could work with something like forwardRef.

```jsx
const AugmentedSandpackPreview = ({ data, ...otherProps }) => {
  const iframeRef =  useRef()

  useEffect(() => {
     iframeRef.current.contentWindow.postMessage(JSON.stringify(data), '*')
  }, [data])

  return <SandpackPreview ref={iframeRef} {...otherProps} />
}
```

## Packages affected

-   [ ] sandpack-client
-   [x] sandpack-react

PR #425, opened 2022-03-27
feat(preview): consume client 

danilowoz:
CodeSandbox logoCodeSandbox logo&nbsp; Open in CodeSandbox <a href="https://codesandbox.io/p/github/codesandbox/sandpack/preview/client">Web Editor</a> | <a href="https://codesandbox.io/p/vscode?owner=codesandbox&repo=sandpack&branch=preview/client">VS Code</a>

Imperatively retrieve the Sandpack client from a SandpackPreview:

```jsx
const SandpackClient: React.FC = () => {
  const previewRef = React.useRef<SandpackPreviewRef>();

  React.useEffect(() => {
    const client = previewRef.current?.getClient();
    const clientId = previewRef.current?.clientId;

    if (client && clientId) {
      console.log(client);
      console.log(sandpack.clients[clientId]);
    }
  }, [sandpack]);

  return <SandpackPreview ref={previewRef} />;
};
```

I agree that currently there is no way to get the current client id and the client instance from a preview component, and it might be too restrictive for a few cases. Although we already have available the list of all clients, currently I can't see a way to link a client from the list to a component, so hope that this new API makes it possible now. 

Worth mentioning, that the SandpackClient will not be instantly available. Sandpack has its own rules to decide what is the "right" moment to initialize a SandpackClient from a preview component (autoron, initMode, client stack), which means that it's expected that `getClient` might return undefined and it's a valid state. 

TODO
- [ ] Tests
- [x] Documentation
- [x] Storybook 

Solves #423 @[malerba118](https://github.com/malerba118) does it make sense for you?

### DIFF
--- a/website/docs/docs/advanced-usage/components.md
+++ b/website/docs/docs/advanced-usage/components.md
@@ -183,7 +183,8 @@ const SandpackPreviewClient: React.FC = () => {
 ```
 
 :::note
-Worth mentioning that the SandpackClient will not be instantly available. Sandpack has its own rules to decide what is the "right" moment to initialize an instance from a preview component (take into account properties such as autorun, initMode, and the current client stack priority), which means that it's expected that `getClient` function returns `undefined` and it's a valid state.
+Worth mentioning that the SandpackClient will not be instantly available. Sandpack has its own rules to decide when it'is the "right" moment to initialize an instance from a preview component. (Sandpack will take into account properties such as autorun, initMode, and the current client stack priority) 
+This means that it's expected that `getClient` function returns `undefined` which is a valid state.
 :::
 
 ## Code Editor
--- a/website/docs/docs/advanced-usage/components.md
+++ b/website/docs/docs/advanced-usage/components.md
@@ -153,7 +153,7 @@ The `<SandpackPreview />` component also allows you to add additional buttons to
 ### Additional content
 For advanced use cases, children of `<SandpackPreview>` are rendered at the end of the preview container.
 
-### Getting client instanceSandpackPreviewRef
+### Getting the client instance from SandpackPreview
 
 You can imperatively retrieve the Sandpack client from a SandpackPreview ref, and also consume or interact with the current state of the preview. Check out the [type definitions](/api/react/interfaces/SandpackPreviewRef) for more details.
 
--- a/website/docs/docs/advanced-usage/components.md
+++ b/website/docs/docs/advanced-usage/components.md
@@ -155,7 +155,7 @@ For advanced use cases, children of `<SandpackPreview>` are rendered at the end 
 
 ### Getting client instanceSandpackPreviewRef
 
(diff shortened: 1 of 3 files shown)


malerba118:
@danilowoz 

Sweet yeah, this looks like it'll work for what i intend to do. 

Thanks for the quick turnaround!

website/docs/docs/advanced-usage/components.md
```diff
@@ -153,6 +153,39 @@ The `<SandpackPreview />` component also allows you to add additional buttons to
 ### Additional content
 For advanced use cases, children of `<SandpackPreview>` are rendered at the end of the preview container.
 
+### Getting client instanceSandpackPreviewRef
```
codebryo, line 156:
```suggestion
### Getting the client instance with SandpackPreviewRef
```

Not sure if that makes sense, but the title did not compute with me considering that the code snippet uses `SandpackPreviewRef`

website/docs/docs/advanced-usage/components.md
```diff
@@ -153,6 +153,39 @@ The `<SandpackPreview />` component also allows you to add additional buttons to
...
+
+You can imperatively retrieve the Sandpack client from a SandpackPreview, consume and interact with the current state of the preview. Check out the [type definitions](/api/react/interfaces/SandpackPreviewRef) for more details.
```
codebryo, line 158:
```suggestion
You can imperatively retrieve the Sandpack client from a SandpackPreview ref, and also consume or interact with the current state of the preview. Check out the [type definitions](/api/react/interfaces/SandpackPreviewRef) for more details.
```

Based on the title, I adjusted the wording here lightly.

website/docs/docs/advanced-usage/components.md
```diff
@@ -153,6 +153,39 @@ The `<SandpackPreview />` component also allows you to add additional buttons to
...
+   * use the whole `sandpack` object as a dependencie.
+   */
+  }, [sandpack]);
+
+  return <SandpackPreview ref={previewRef} />;
+};
+```
+
+:::note
+Worth mentioning that the SandpackClient will not be instantly available. Sandpack has its own rules to decide what is the "right" moment to initialize an instance from a preview component (take into account properties such as autorun, initMode, and the current client stack priority), which means that it's expected that `getClient` function returns `undefined` and it's a valid state.
```
codebryo, line 186:
```suggestion
Worth mentioning that the SandpackClient will not be instantly available. Sandpack has its own rules to decide when it'is the "right" moment to initialize an instance from a preview component. (Sandpack will take into account properties such as autorun, initMode, and the current client stack priority) 
This means that it's expected that `getClient` function returns `undefined` which is a valid state.
```

codebryo (commented):
Hope the changes make sense. Good job Danilo

alexnm (approved):

website/docs/docs/advanced-usage/components.md
danilowoz:
```suggestion
### Getting the client instance from SandpackPreview
```

codebryo: Reads good!

PR #460, opened 2022-05-06
Upgrade React template to 18

gaearon:
This would need to go in a major probably? Since it changes existing sandboxes to 18.


danilowoz:
Thanks @gaearon! I applied your changes to #375, which is going to be a major release 

https://github.com/codesandbox/sandpack/pull/375/commits/e48e659582944a2c2a0148dc78e73bfc6e40aeaa

PR #461, opened 2022-05-06
fix(editor): prefer useId when available

gaearon:
Haven't actually tested but we need to do this. I see mismatches otherwise.

I tried to keep it backwards-compatible. If you drop 17 support then you can just useId.

### DIFF
--- a/sandpack-react/src/components/CodeEditor/CodeMirror.tsx
+++ b/sandpack-react/src/components/CodeEditor/CodeMirror.tsx
@@ -33,6 +33,7 @@ import { getFileName } from "../../utils/stringUtils";
 
 import { highlightDecorators } from "./highlightDecorators";
 import { highlightInlineError } from "./highlightInlineError";
+import { useGeneratedId } from "./useGeneratedId";
 import {
   getCodeMirrorLanguage,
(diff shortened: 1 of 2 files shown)


sandpack-react/src/components/CodeEditor/CodeMirror.tsx
```diff
@@ -125,7 +125,10 @@ export const CodeMirror = React.forwardRef<CodeMirrorRef, CodeMirrorProps>(
 
     const c = useClasser("sp");
     const { listen } = useSandpack();
-    const ariaId = React.useRef<string>(id ?? generateRandomId());
```
gaearon, line 128: Note sure if the "passed-in ID" use case is important.

danilowoz: Initially, the purpose of the `id` props is to get consistent ids in snapshot tests. But I guess `useId` will always return a constant value, right?

gaearon: As long as you don't change the parent component path. If you do, the snapshot will update.

PR #474, opened 2022-05-26
fix(codemirror): remove inline props

danilowoz:
CodeSandbox logoCodeSandbox logo&nbsp; Open in CodeSandbox <a href="https://codesandbox.io/p/github/codesandbox/sandpack/fix/extension-inline-prop">Web Editor</a> | <a href="https://codesandbox.io/p/vscode?owner=codesandbox&repo=sandpack&branch=fix/extension-inline-prop">VS Code</a>

Address https://github.com/reactjs/reactjs.org/issues/4686#issuecomment-1137402613

All `useEffect` which depends on inline default props (`extensions` and `extensionsKeymap`), were constantly being triggered unnecessarily due to these props always returning a new value every time that the parent rerender.  

@gaearon does it make sense? 

----

Note to me: apply these changes on `main` too.

### DIFF
--- a/sandpack-react/src/utils/array.ts
+++ b/sandpack-react/src/utils/array.ts
@@ -3,5 +3,15 @@
 export const shallowEqual = (a: any[], b: any[]): boolean => {
   if (a.length !== b.length) return false;
 
-  return a.every((item, index) => item === b[index]);
+  let result = true;
+
+  for (let index = 0; index < a.length; index++) {
+    if (a[index] !== b[index]) {
+      result = false;
+
+      break;
+    }
+  }
+
+  return result;
 };
--- a/sandpack-react/src/utils/array.ts
+++ b/sandpack-react/src/utils/array.ts
@@ -1,0 +1,7 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+export const shallowEqual = (a: any[], b: any[]): boolean => {
+  if (a.length !== b.length) return false;
+
+  return a.every((item, index) => item === b[index]);
+};
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
   },
   "resolutions": {
     "react": "18.1.0",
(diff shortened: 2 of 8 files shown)


gaearon:
What if the user specifies an inline value? Like `extensions={[onLint]}`. I think you 
might need to diff them manually inside the effect instead.

gaearon (approved):

danilowoz:
Indeed there is a significant different between them: 
Screenshot 2022-05-26 at 16 53 34

sandpack-react/src/utils/array.ts
```diff
@@ -0,0 +1,7 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+export const shallowEqual = (a: any[], b: any[]): boolean => {
+  if (a.length !== b.length) return false;
+
+  return a.every((item, index) => item === b[index]);
```
gaearon, line 6: Nit: this always traverses all the items. A regular for loop would be able to exit early on the first mismatches. Doesn't matter for small arrays though.

gaearon: Actually wait, I'm wrong. I think it _would_ exit early. So nvm.

danilowoz: If I'm doing this right, for-loop is still the fastest option compared to the `some` or `every` implementation: 
Screenshot 2022-05-26 at 17 33 08

danilowoz: As far as I know, `some` or `every` exit as soon as a predicate match, but I couldn't find this in any documentation. If you have it, please let me know

gaearon: loops ftw

sophiebits: Yes, it does an early exit ("If _testResult_ is **false**, return **false**."). https://tc39.es/ecma262/#sec-array.prototype.every

PR #527, opened 2022-07-08
fix(codemirror): use correct method to indent the code on tab key press

### LINKED ISSUE #522: “extensionsKeymap” doesn't let us redefine existing keymaps
# Bug report

## Packages affected

-   [ ] sandpack-client
-   [x] sandpack-react

## Description of the problem

A few of my users have noticed that the tab-handling is a bit funny: "Tab" inserts a literal tab character (`\t`), but Shift + Tab removes 2 spaces.

I'd prefer to have Tab always add 2 spaces when no text is selected, so I updated my code to include a "Tab" keymap:

```js
import { indentMore } from '@codemirror/commands';

<SandpackCodeEditor
  extensionsKeymap={[
    {
      key: 'Tab',
      run: ({ state, dispatch }) => {
        console.log('Test')
        return true;
      },
    },
  ]}
/>
```

My actual code tried to perform the tab operation, but I've simplified it here to the minimal code required to demonstrate the issue. When I hit "Tab", callback isn't called, and so nothing I do in this function works.

### What steps can we take to reproduce the problem?

Check out the linked CodeSandbox. It defines keymaps for the lowercase "a" character and Tab, removing their default functionality and logging a console message.

It works for "a", I believe because there is no conflicting override. But it doesn't work for "Tab".

I wonder if the issue has to do with [keymap.of](https://github.com/codesandbox/sandpack/blob/main/sandpack-react/src/components/CodeEditor/CodeMirror.tsx#L229); I wonder if it isn't merging them in the order provided. I tried digging into the codemirror code but couldn't find this method defined aaywhere.

### Link to sandbox: 

[**CodeSandbox Repro**](https://codesandbox.io/s/sandpack-extensionskeymap-bug-lyh039?file=/src/App.js)

### Your Environment

| Sandpack-client version |   1.1.0    |
| Sandpack-react version  | 1.1.1 |
| Browser                 |    Chrome  |
| Operating System        |   MacOS |

Thanks y'all 💖 Really appreciate the work you've been doing on this project!

danilowoz:
CodeSandbox logoCodeSandbox logo&nbsp; Open in CodeSandbox <a href="https://codesandbox.io/p/github/codesandbox/sandpack/draft/dank-surf">Web Editor</a> | <a href="https://codesandbox.io/p/vscode?owner=codesandbox&repo=sandpack&branch=draft/dank-surf">VS Code</a>

Hey @joshwcomeau, this PR addresses your bug report about the ability to override the `extensionsKeymap`. Turns out, it was a mistake on our side and we were literally inserting a `tab` instead of regularly indenting the line. So, this should fix your original problem and make the CodeEditor behavior more consistent.

Closes #522

### DIFF
--- a/sandpack-react/src/components/CodeEditor/CodeMirror.tsx
+++ b/sandpack-react/src/components/CodeEditor/CodeMirror.tsx
@@ -4,8 +4,8 @@ import { closeBrackets, closeBracketsKeymap } from "@codemirror/closebrackets";
 import {
   defaultKeymap,
   indentLess,
+  indentMore,
   deleteGroupBackward,
-  insertTab,
 } from "@codemirror/commands";
 import { commentKeymap } from "@codemirror/comment";
 import { lineNumbers } from "@codemirror/gutter";
@@ -100,7 +100,7 @@ interface CodeMirrorProps {
    */
   id?: string;
   extensions?: Extension[];
-  extensionsKeymap?: Array<readonly KeyBinding[]>;
+  extensionsKeymap?: KeyBinding[];
 }
 
 export interface CodeMirrorRef {
@@ -197,7 +197,7 @@ export const CodeMirror = React.forwardRef<CodeMirrorRef, CodeMirrorProps>(
         const customCommandsKeymap: KeyBinding[] = [
           {
             key: "Tab",
-            run: insertTab,
+            run: indentMore,
           },
           {
             key: "Shift-Tab",
--- a/sandpack-react/src/components/CodeEditor/CodeMirror.tsx
+++ b/sandpack-react/src/components/CodeEditor/CodeMirror.tsx
@@ -197,10 +197,14 @@ export const CodeMirror = React.forwardRef<CodeMirrorRef, CodeMirrorProps>(
         const customCommandsKeymap: KeyBinding[] = [
           {
             key: "Tab",
-            run: ({ state, dispatch }): boolean => {
-              indentMore({ state, dispatch });
+            run: (view): boolean => {
+              indentMore(view);
 
-              return false;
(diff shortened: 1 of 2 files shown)


sandpack-react/src/components/CodeEditor/CodeMirror.tsx
```diff
@@ -197,11 +197,19 @@ export const CodeMirror = React.forwardRef<CodeMirrorRef, CodeMirrorProps>(
         const customCommandsKeymap: KeyBinding[] = [
           {
             key: "Tab",
-            run: insertTab,
+            run: ({ state, dispatch }): boolean => {
+              indentMore({ state, dispatch });
+
+              return false;
```
danilowoz, line 203: As `indentLess` returns true, this was preventing running the next command. So by returning false, it's possible to run multiple commands on the key press. 

I'm not absolutely sure about this change, but on my tests, it didn't introduce any regression,

sandpack-react/src/components/CodeEditor/CodeMirror.tsx
```diff
@@ -197,11 +197,27 @@ export const CodeMirror = React.forwardRef<CodeMirrorRef, CodeMirrorProps>(
         const customCommandsKeymap: KeyBinding[] = [
           {
             key: "Tab",
-            run: insertTab,
+            run: (view): boolean => {
+              indentMore(view);
```
danilowoz, line 201: As `indentMore` returns true, this was preventing running the next command. So, here I'm grabbing the custom keybinding and running before returning it.

I'm not fully confident, but it seems to work

joshwcomeau:
@danilowoz Woohoo! Thanks for looking into this 😄 Really appreciate it!

PR #542, opened 2022-07-26
feat(custom-setup): introduce custom npm registries

danilowoz:
CodeSandbox logoCodeSandbox logo&nbsp; Open in CodeSandbox <a href="https://codesandbox.io/p/github/codesandbox/sandpack/draft/goofy-leftpad">Web Editor</a> | <a href="https://codesandbox.io/p/vscode?owner=codesandbox&repo=sandpack&branch=draft/goofy-leftpad">VS Code</a>

Documentation preview: https://1gemwv-3000.preview.csb.app/guides/private-packages
Screenshot 2022-07-26 at 15 56 58


examples/custom-npm-registry/README.md
```diff
@@ -0,0 +1,24 @@
+# Sandpack custom npm registry (proxy)
+
+This project works as a recipe to proxy NPM registries. The main goal is to consume private packages and expose them through a private registry without any authentication token (GitHub or Npm) or requiring any authentication process, 
+
+This new private registry can be used on a Sandpack instance and run private packages public. 
+
+Disclaimer: it would be best if you kept in mind that it will expose all private packages in your account, so be careful where this proxy will be used and the scopes of the authentication token. That's why we initially set this project for read-only purposes, meaning no user would be able to publish changes using this proxy.
```
DeMoorJasper: It is possible for users to define only certain packages to expose, if customScopes is `@scope/package-name` instead of `@scope/*` it will only expose that one package, they can even do something like `@scope/design-system*` to expose all packages of the design system or something

examples/custom-npm-registry/htpasswd
```diff
@@ -0,0 +1 @@
+test:{SHA}qUqP5cyxm6YcTAhz05Hph5gvu9M=
```
DeMoorJasper, line 1: Maybe add a comment mentioning this is only a test user with password `test`

website/docs/docs/faq.md
```diff
@@ -9,9 +9,13 @@ title: FAQ
 
 Sandpack is an open ecosystem of components and utilities that allow you to compile and run modern frameworks in the browser. You can either use one of our predefined `components` for embedding the *CodeSandbox* experience into your projects, or you can build your own version of `sandpack`, on top of our standard components and utilities.
 
-#### How to load local or private dependencies?
+#### How to load private dependencies?
 
-Currently, Sandpack doesn’t have a way to consume private dependencies from any kind of registry service, because the bundler host is shared with all Sandpack consumers apps. However, you can pass local dependencies just like a regular file or using the external resource API:
+Read the following [guide](/guides/private-packages).
+
+#### How to load local dependencies?
```
DeMoorJasper, line 16: ESModules might also be a solution for local dependencies, as we support importing files as a full url. although it never really took of in codesandbox itself, don't even think we ever even announced that

danilowoz: humm good to know. I think I'm going to work on local dependencies in another PR, where we could explore how Verdaccio can also help in this regard.

website/docs/docs/guides/private-packages.md
```diff
@@ -0,0 +1,64 @@
+---
+sidebar_position: 2
+title: Private packages
+---
+
+<big>The custom private npm registry allows Sandpack instances to retrieve npm packages from your npm registry. This option requires running a third service (Node.js server) and configuring your Sandpack provider to consume these dependencies from another registry, not the public ones.</big>
```
danilowoz, line 6:
```suggestion
<big>The custom private NPM registry allows Sandpack instances to retrieve private NPM packages from your own registry. This option requires running a third service (Node.js server) and configuring your Sandpack provider to consume these dependencies from another registry, not the public ones.</big>
```

website/docs/docs/guides/private-packages.md
```diff
@@ -0,0 +1,64 @@
...
+`,
+    }}
+    template="react"
+  />
+```
+
+## Security
+It's essential to keep the information and tokens of the npm registry private, so by using this method, it would be best if you kept in mind that it might expose all private packages in your account, so be careful where and how this proxy will be used. Plus, make sure only to use authentication tokens with read-only access.
+
+It's also possible to define only certain packages to be exposed. If custom scopes are `@scope/package-name` instead of `@scope/*`, it will only expose that one package. You can even do something like `@scope/design-system*` to expose all packages of the design system or something.
```
codebryo, line 63:
```suggestion
It's also possible to expose only specific packages. If the custom scopes are `@scope/package-name` instead of `@scope/*`, it will only expose that particular package. You can even do something like `@scope/design-system*` to expose all packages of the design system.

```

codebryo (commented):
Added my suggestion for documentation changes.

PR #570, opened 2022-09-08
feat(codeeditor): support additional, user-supplied language modes

jodyheavener:

Feature

Sandpack language support is currently limited to the [built-in languages](https://github.com/codesandbox/sandpack/blob/main/sandpack-react/src/components/CodeEditor/utils.ts#L155-L160). This means that you cannot currently use Sandpack with code in other languages correctly (it would [default to Markdown](https://github.com/codesandbox/sandpack/blob/main/sandpack-react/src/components/CodeEditor/utils.ts#L192-L193)).

## What is the new behavior?

This PR adds the ability to add custom language modes by supplying a language type, applicable file extensions, and a `LanguageSupport` instance for that syntax mode.

The `CodeEditor` component now takes an `additionalLanguages` prop of the following type:

```ts
Array<{
  name: string;
  extensions: string[];
  language: LanguageSupport;
}>
```

(The `Sandpack` preset component also takes this in via its `options.codeEditor` prop object and passes it down to `CodeEditor`.)

This new prop value is pulled out of the context in `CodeMirror` and passed down to the `getLanguageFromFile` to establish the language name, and `getCodeMirrorLanguage` to extract the custom language support and pass it over to CodeMirror like our other language extensions.

The following Storybook entries show off the new functionality:

- `/story/presets-sandpack-options--custom-languages` shows off the usage of multiple custom languages in a preset
- `/story/components-codemirror--custom-language-shell` shows off the usage of the custom Shell language
- `/story/components-codemirror--custom-language-python` shows off the usage of the custom Python language

## Checklist

- [x] Documentation;
- [ ] Tests;
- [ ] Ready to be merged;

## Additional comments

This is mostly a proof of concept, and I welcome any larger structural feedback! I also have not written tests yet; I will do so if this approach seems correct.

One thing I have noticed is that this approach requires the loosening of language types, specifically `CodeMirrorProps["fileType"]` and `getLanguageFromFile`'s return type of `SandpackLanguageSupport`. I believe this is necessary for the support of custom languages, but it's something to keep in mind when reviewing - perhaps there is a better approach.


sandpack-react/src/components/CodeEditor/CodeMirror.stories.tsx
```diff
@@ -95,6 +99,46 @@ export const Markdown: React.FC = () => (
...
+      id="shell"
+      initMode="immediate"
+      showLineNumbers={false}
+    />
+  </SandpackProvider>
+);
+
+export const CustomLanguagePython: React.FC = () => (
+  <SandpackProvider
+    additionalLanguages={[
```
danilowoz, line 124: I think you can pass this prop directly to CodeEditor instead of SandpackProvider because this kind of capability doesn't need to be shared between other components, it only matters for the CodeMirror instance.

sandpack-react/src/components/CodeEditor/CodeMirror.tsx
```diff
@@ -65,17 +65,7 @@ export type Decorators = Array<{
...
-    | "jsx"
-    | "ts"
-    | "tsx"
-    | "css"
-    | "scss"
-    | "less"
-    | "html"
-    | "vue"
-    | "markdown";
+  fileType?: string;
```
danilowoz, line 68: I'm wondering if we can extend this key using generics and still keep type-safe, let me know if you need any help

sandpack-react/src/components/CodeEditor/__snapshots__/useSyntaxHighlight.test.tsx.snap
```diff
@@ -1456,6 +1456,21 @@ Array [
...
+api_token = os.environ",
+  <span
+    className="ͼo"
+  >
+    ['API_TOKEN']
+  </span>,
+]
+`;
+
+exports[`useSyntaxHighlight renders a shell block 1`] = `null`;
```
danilowoz, line 1472: Interesting why the output is null, could you take a look at this?

sandpack-react/src/presets/Sandpack.stories.tsx
```diff
@@ -286,3 +290,39 @@ export const ShowConsoleButton: React.FC = () => (
     />
   </div>
 );
+
+export const CustomLanguages: React.FC = () => (
+  <Sandpack
+    additionalLanguages={[
```
danilowoz, line 296: I'd add the `additionalLanguages` inside of `options. codeEditor`, which would make clearer where this key belongs

sandpack-react/src/presets/Sandpack.stories.tsx
```diff
@@ -286,3 +290,39 @@ export const ShowConsoleButton: React.FC = () => (
...
+      {
+        name: "python",
+        extensions: ["py"],
+        language: new LanguageSupport(StreamLanguage.define(python)),
+      },
+      {
+        name: "shell",
+        extensions: ["sh"],
```
danilowoz, line 304: Could you add more extension examples? This structure looks good, btw!

sandpack-react/src/components/CodeEditor/CodeMirror.tsx
```diff
@@ -144,7 +134,10 @@ export const CodeMirror = React.forwardRef<CodeMirrorRef, CodeMirrorProps>(
     );
 
     const c = useClasser(THEME_PREFIX);
-    const { listen } = useSandpack();
+    const {
+      listen,
+      sandpack: { additionalLanguages },
```
danilowoz, line 139: I believe we can get this value from this component `props`, instead of the context. The additional language syntax doesn't need to be widely shared between sandpack instances, as this is a very particular implementation of CodeMirror.

danilowoz (approved):
Thanks @jodyheavener, for taking this issue! I'm glad it works, and I just left some comments about how you're consuming the props, great job!

sandpack-react/src/components/CodeEditor/CodeMirror.stories.tsx
jodyheavener: Oh yes this definitely simplified things 👌

sandpack-react/src/components/CodeEditor/__snapshots__/useSyntaxHighlight.test.tsx.snap
jodyheavener: This is odd - it looks like it's stumbling on the provided code sample? If I change the mocked "shell" code to the Python code it works fine.

jodyheavener: Was exploring this a bit more, and it looks like this occurs with Markdown as well: https://github.com/codesandbox/sandpack/blob/main/sandpack-react/src/components/CodeEditor/__snapshots__/useSyntaxHighlight.test.tsx.snap#L1457

sandpack-react/src/components/CodeEditor/CodeMirror.tsx
jodyheavener: I've been wracking my brain trying to make this work, but I'm not sure I'm fully understanding how we can use generics in this case. Would love some guidance if you have a chance!

sandpack-react/src/components/CodeEditor/CodeMirror.stories.tsx
```diff
@@ -123,13 +123,13 @@ export const CustomLanguagePython: React.FC = () => (
     <CodeEditor
       additionalLanguages={[
         {
-          name: "python",
+          name: "python" as const, // TODO: how to convert it to a union type instead of a string?
```
jodyheavener, line 126: This is a cool approach! Does it need to be a union, since it would never be concerned about existing file types?

PR #581, opened 2022-09-27
fix(codemirror): avoid Escape keybinding conflict

danilowoz:
Solves https://github.com/codesandbox/sandpack/discussions/576
Solves https://github.com/codesandbox/sandpack/issues/582


danilowoz:
Thanks @hawkup, for providing an example. I managed to fix it by prioritizing the extensions keybinding over the internal ones, could you please test it and make sure it's working as you expected? 

https://codesandbox.io/s/sandpack-with-codemirror-vim-7g465c

hawkup:
@danilowoz it works like my expectation. Thank you so much.

PR #601, opened 2022-10-02
fix: style prop no longer passed to React.Fragment

CTNicholas:

Bug fix

Currently getting this error when using Sandpack React:
```
Invalid prop `style` supplied to `React.Fragment`. React.Fragment can only have `key` and `children` props.
```

I traced back the error back to this point in the code.

## What is the new behavior?

Don't apply the `style` prop to `React.Fragment`

It seemed fine in CodeSandbox projects, but I have done no more testing. Please check everything works!

### DIFF
--- a/sandpack-react/src/presets/Sandpack.tsx
+++ b/sandpack-react/src/presets/Sandpack.tsx
@@ -82,6 +82,7 @@ export const Sandpack: SandpackInternal = (props) => {
 
   const hasRightColumn =
     props.options?.showConsole || props.options?.showConsoleButton;
+  const RightColumn = hasRightColumn ? SandpackStack : React.Fragment;
 
   const rightColumnStyle = {
     flexGrow: previewPart,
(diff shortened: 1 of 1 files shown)


sandpack-react/src/presets/Sandpack.tsx
```diff
@@ -93,6 +92,13 @@ export const Sandpack: SandpackInternal = (props) => {
     height: props.options?.editorHeight, // use the original editor height
   };
 
+  const RightColumn: React.FC<{ children: React.ReactNode }> = ({ children }) =>
```
danilowoz, line 95: Hummm... I'm afraid that this will remount everything and its children on every render. 

What about conditionally passing the props to the `RightColumn`? Something like this:

```jsx
const rightColumnProps = useMemo(() => hasRightColumn ? { style } : {}, [hasRightColumn])

<RightColumn {...rightColumnProps}>
```

CTNicholas: Wow yep, I wasn't thinking, haha. Updated 👍

PR #644, opened 2022-11-19
chore(docs): Add `key` to SSR snippets

LekoArts:

Small bug fix

Gatsby complains with:

```shell
Warning: Each child in a list should have a unique "key" prop.
```

## What is the new behavior?

Adding a `key` to the style tags ensures that React has a non-null value for `key`.

Running a Gatsby site locally with that script

## Checklist

- [x] Documentation;
- [ ] Tests;
- [x] Ready to be merged;


danilowoz:
Thanks

PR #659, opened 2022-12-12
fix(sandpack-react): safeguard against full page crash when there is no visible files

AaronPowell96:

Bux fix? 

TLDR: Added lots of optional chaining on "visibleFiles" and "code" to prevent errors where code is undefined or there are no visibleFiles.

This change prevents sandpack from erroring out in the scenario where there a no "active files" in view, which results in "code" being undefined among other things.

While this behaviour is not possible unless you start to play around with the 'deleteFile' consumer from context, I do believe  this would be a great "fix" to get in for people to play around further with customizing sandpack.
To clarify, this "bug" is not possible unless you use the sandpack provider and delete the active file with no other files being open.

image

Currently if you delete the last remaining active file, the whole page crashes.

## What is the new behavior?

New behaviour, the page doesn't crash and you receive a relevant error inside of the sandbox. This seems correct. Seen below (Deleted app.ts from the react.ts template). 
image

With the following console errors:
image

1. Create a local sandpack
2. Use the deleteFile from context on the only active file
3. Boom

## Checklist

- [ N/A ] Documentation;
- [ N/A ] Storybook (if applicable);
- [ N/A ] Tests;
- [X] Ready to be merged;

### DIFF
--- a/sandpack-react/src/components/CodeEditor/CodeMirror.tsx
+++ b/sandpack-react/src/components/CodeEditor/CodeMirror.tsx
@@ -389,11 +389,12 @@ export const CodeMirror = React.forwardRef<CodeMirrorRef, CodeMirrorProps>(
     React.useEffect(() => {
       if (cmView.current && code !== internalCode) {
         const view = cmView.current;
-        const selection = view.state.selection.ranges.some(
-          ({ to, from }) => to > code?.length || from > code?.length
-        )
-          ? EditorSelection.cursor(code?.length)
-          : view.state.selection;
+        const selection =
+          view.state.selection.ranges.some(
+            ({ to, from }) => to > code?.length || from > code?.length
+          ) && code !== undefined
+            ? EditorSelection.cursor(code?.length)
+            : view.state.selection;
 
         const changes = {
           from: 0,
--- a/sandpack-react/src/components/CodeEditor/CodeMirror.tsx
+++ b/sandpack-react/src/components/CodeEditor/CodeMirror.tsx
@@ -389,16 +389,22 @@ export const CodeMirror = React.forwardRef<CodeMirrorRef, CodeMirrorProps>(
     React.useEffect(() => {
       if (cmView.current && code !== internalCode) {
         const view = cmView.current;
-
         const selection = view.state.selection.ranges.some(
-          ({ to, from }) => to > code.length || from > code.length
+          ({ to, from }) => to > code?.length || from > code?.length
         )
-          ? EditorSelection.cursor(code.length)
+          ? EditorSelection.cursor(code?.length)
           : view.state.selection;
 
-        const changes = { from: 0, to: view.state.doc.length, insert: code };
+        const changes = {
+          from: 0,
+          to: view.state.doc?.length,
+          insert: code ?? "",
+        };
 
-        view.dispatch({ changes, selection });
+        view.dispatch({
+          changes,
+          selection,
+        });
       }
       // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [code]);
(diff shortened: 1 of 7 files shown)


sandpack-react/src/components/CodeEditor/CodeMirror.tsx
```diff
@@ -389,16 +389,23 @@ export const CodeMirror = React.forwardRef<CodeMirrorRef, CodeMirrorProps>(
...
-          ? EditorSelection.cursor(code.length)
-          : view.state.selection;
-
-        const changes = { from: 0, to: view.state.doc.length, insert: code };
-
-        view.dispatch({ changes, selection });
+        const selection =
+          view.state.selection.ranges.some(
+            ({ to, from }) => to > code?.length || from > code?.length
+          ) && code !== undefined
```
AaronPowell96, line 395: This additional check means that when the only active file is deleted, there is no leftover code in the codeEditor.

sandpack-react/src/components/CodeEditor/CodeMirror.tsx
```diff
@@ -389,16 +389,23 @@ export const CodeMirror = React.forwardRef<CodeMirrorRef, CodeMirrorProps>(
...
+            ? EditorSelection.cursor(code?.length)
+            : view.state.selection;
+
+        const changes = {
+          from: 0,
+          to: view.state.doc?.length,
+          insert: code ?? "",
```
AaronPowell96, line 402: Inserting undefined created some errors with a CodeMirror plugin

danilowoz (commented):
From the product point of view, I think it would be more interesting always show a file instead of an empty state. On delete, we could fall back to the next visible file, entry file, or the next file in the state. What do you think? 

That said, I think your changes in this PR are still valid, but I'm also interested in how it would look for the final users.

sandpack-react/src/components/CodeEditor/index.tsx
```diff
@@ -102,7 +102,7 @@ export const SandpackCodeEditor = React.forwardRef<
     const { sandpack } = useSandpack();
     const { code, updateCode, readOnly: readOnlyFile } = useActiveCode();
     const { activeFile, status, editorState } = sandpack;
-    const shouldShowTabs = showTabs ?? sandpack.visibleFiles.length > 1;
+    const shouldShowTabs = showTabs ?? sandpack.visibleFiles?.length > 1;
```
danilowoz, line 105: When and why `visibleFiles` becomes undefined? It should always return an array.

AaronPowell96:
> From the product point of view, I think it would be more interesting always show a file instead of an empty state. On delete, we could fall back to the next visible file, entry file, or the next file in the state. What do you think?
> 
> That said, I think your changes in this PR are still valid, but I'm also interested in how it would look for the final users.

I think this is an interesting approach too. When deleting the final activeFile having a codeEditor available with no code to edit / no files being changed did feel a little inconsistent.

I'll try to look into this today and raise an alternative PR.

sandpack-react/src/components/CodeEditor/CodeMirror.tsx
AaronPowell96: iirc I attempted this and it caused errors. I like your idea of falling back to the entry file and will attempt that first. Cheers for the feedback!

sandpack-react/src/components/CodeEditor/index.tsx
AaronPowell96: > From the product point of view, I think it would be more interesting always show a file instead of an empty state. On delete, we could fall back to the next visible file, entry file, or the next file in the state. What do you think?
> 
> That said, I think your changes in this PR are still valid, but I'm also interested in how it would look for the final users.

I've reached out on twitter with a few questions, let me know if you'd prefer a github discussion etc!

PR #661, opened 2022-12-12
fix(sandpack-react): prevent crash upon deleting last visible file

AaronPowell96:

Alternative to https://github.com/codesandbox/sandpack/pull/659

TLDR: When deleteFile is called, if there are no visibleFiles then open the entryFile. Also set the activeFile is the currently activeFile is deleted.

This change prevents sandpack from erroring out in the scenario where there a no "active files" in view, which results in "code" being undefined among other things.

image

Currently if you delete the last remaining active file, the whole page crashes.

## What is the new behavior?

New behaviour, the page doesn't crash and you receive a relevant error inside of the sandbox. This seems correct. Seen below (Deleted app.ts from the react.ts template). 
This automatically opens the entry file and presents an error related to user code.

image

1. Create a local sandpack
2. Use the deleteFile from context on the active file
3. Boom

## Checklist

- [ N/A ] Documentation;
- [ N/A ] Storybook (if applicable);
- [ N/A ] Tests;
- [X] Ready to be merged;

### DIFF
--- a/sandpack-react/src/contexts/sandpackContext.test.tsx
+++ b/sandpack-react/src/contexts/sandpackContext.test.tsx
@@ -54,6 +54,38 @@ describe(SandpackProvider, () => {
       ]);
     });
 
+    it("deletes the activeFile and set the following visibleFile as active", () => {
+      const root = create(
+        <SandpackProvider
+          template="react"
+          options={{ activeFile: "/App.js", visibleFiles: ["/styles.css"] }}
+        />
+      );
+
+      const instance = root.getInstance();
+
+      instance.runSandpack();
+      instance.deleteFile("/App.js");
+
+      expect(instance.state.activeFile).toBe("/styles.css");
+    });
+
+    it("deletes the activeFile and set the entry file if there no visibleFile left", () => {
+      const root = create(
+        <SandpackProvider
+          template="react"
+          options={{ activeFile: "/App.js", visibleFiles: [] }}
+        />
+      );
+
+      const instance = root.getInstance();
+
+      instance.runSandpack();
+      instance.deleteFile("/App.js");
+
+      expect(instance.state.activeFile).toBe("/index.js");
+    });
+
     it("updates a file", () => {
       const instance = createContext();
 
@@ -105,7 +137,7 @@ describe(SandpackProvider, () => {
       expect(instance.state.editorState).toBe("dirty");
     });
 
-    it("should return a pristine value after reseting files", () => {
+    it("should return a pristine value after reset files", () => {
       const instance = createContext();
 
       expect(instance.state.editorState).toBe("pristine");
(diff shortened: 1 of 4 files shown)


AaronPowell96:
@danilowoz Do I need to do anything here or is this change reliant on a team member at the moment? Cheers! Not sure if it's waiting for me to do something

danilowoz:
Hey, I'll take a look at this by tomorrow, but in general it looks good to me

Thanks 😄

sandpack-react/src/contexts/sandpackContext.tsx
```diff
@@ -519,12 +519,31 @@ export class SandpackProviderClass extends React.PureComponent<
...
       const newFiles = { ...files };
       delete newFiles[path];
 
+      const remainingVisibleFiles = visibleFiles.filter(
+        (openPath) => openPath !== path
+      );
+      const deletedLastVisibleFile = remainingVisibleFiles.length === 0;
+
+      if (deletedLastVisibleFile) {
+        const nextFile = Object.keys(files)[Object.keys(files).length - 1];
```
danilowoz, line 532: @AaronPowell96 I made a few changes to make the next available file instead of the entry file. I believe it's the safest way to go because I might need to consider that any file can be deleted, even the entry file.

What do you think?

AaronPowell96: Okay that's better in my opinion too! Nice work

sandpack-react/src/contexts/sandpackContext.test.tsx
```diff
@@ -54,6 +54,38 @@ describe(SandpackProvider, () => {
...
+
+      const instance = root.getInstance();
+
+      instance.runSandpack();
+      instance.deleteFile("/App.js");
+
+      expect(instance.state.activeFile).toBe("/styles.css");
+    });
+
+    it("deletes the activeFile and set the entry file if there no visibleFile left", () => {
```
AaronPowell96, line 73: With your change does this test need to be modified as we don't expect just index.js

sandpack-react/src/contexts/sandpackContext.test.tsx
```diff
@@ -54,6 +54,38 @@ describe(SandpackProvider, () => {
...
+          options={{ activeFile: "/App.js", visibleFiles: [] }}
+        />
+      );
+
+      const instance = root.getInstance();
+
+      instance.runSandpack();
+      instance.deleteFile("/App.js");
+
+      expect(instance.state.activeFile).toBe("/index.js");
```
danilowoz, line 86:
```suggestion
      expect(instance.state.activeFile).toBe("/package.json");
```

danilowoz:
Thank you @AaronPowell96

PR #667, opened 2022-12-15
fix: add predefined template

jerrywu001:
```diff
/**
 * @category Setup
 */
export type SandpackPredefinedTemplate =
  | "angular"
  | "react"
  | "react-ts"
  | "test-ts"
  | "vanilla"
  | "vanilla-ts"
  | "vue"
  | "vue3"
+  | "vue3-ts"
  | "svelte"
  | "solid";
```

### DIFF
--- a/sandpack-react/src/types.ts
+++ b/sandpack-react/src/types.ts
@@ -241,6 +241,7 @@ export type SandpackPredefinedTemplate =
   | "vanilla-ts"
   | "vue"
   | "vue3"
+  | "vue3-ts"
   | "svelte"
   | "solid";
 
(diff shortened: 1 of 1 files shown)


sandpack-react/src/types.ts
```diff
@@ -241,6 +241,7 @@ export type SandpackPredefinedTemplate =
   | "vanilla-ts"
   | "vue"
   | "vue3"
+  | "vue3-ts"
```
danilowoz, line 244: I was wondering if we could make this smarter instead of a manual list of types?
```
export type SandpackPredefinedTemplate = keyof typeof SANDBOX_TEMPLATES;
 ```

What do you think?

jerrywu001: 👍🏻
I have updated.

danilowoz (approved):
Thank you!

PR #675, opened 2023-01-04
feat(Sandpack-preset): split panel

danilowoz:
This introduces the ability to resize the columns and rows on `<Sandpack />` (preset component):
- Desktop only: which means that mobile or touch devices don't support resizing panels in order to keep the usability accessible as the drag handler is not visible for non-cursor pointers;
- Boundaries: columns and rows have a minimum of 25% size and a max of 75%;

Plus, this PR introduces other improvements which aim to solve some layout issues on small-screen devices (set min-height for standard components), and other things to reduce the bundle size. 

Todo
- [x] Docs: remove handler for nested Sandpack layout;
- [x] Docs: document `resizablePanels` prop;
- [x] Rows handler: hide handler when vertical panel is collapsed (console panel is hidden);

Preview: https://2075ed-6006.preview.csb.app/iframe.html?id=intro-playground--preset&args=&viewMode=story

### DIFF
--- a/sandpack-react/src/components/common/Layout.tsx
+++ b/sandpack-react/src/components/common/Layout.tsx
@@ -33,7 +33,6 @@ export const layoutClassName = css({
     flexGrow: 1,
     flexShrink: 1,
     flexBasis: "0",
-    minWidth: "350px",
     height: "$layout$height",
 
     "@media print": {
--- a/sandpack-react/src/components/common/Layout.tsx
+++ b/sandpack-react/src/components/common/Layout.tsx
@@ -41,10 +41,13 @@ export const layoutClassName = css({
     },
 
     "@media screen and (max-width: 768px)": {
-      height: "auto",
+      [`&:not(.${THEME_PREFIX}-preview, .${THEME_PREFIX}-editor, .${THEME_PREFIX}-preset-column)`]:
+        {
+          height: "calc($layout$height / 2)",
+        },
 
       /* triggers the layout break at the 768px breakpoint, not when the component is less then 700px */
-      minWidth: "100% !important;",
+      minWidth: "100%",
     },
   },
 
--- a/sandpack-react/src/components/common/Stack.tsx
+++ b/sandpack-react/src/components/common/Stack.tsx
@@ -13,7 +13,6 @@ export const stackClassName = css({
   width: "100%",
   position: "relative",
   backgroundColor: "$colors$surface1",
-  // transition: "flex $transitions$default",
   gap: 1, // border between components
 
   [`&:has(.${THEME_PREFIX}-stack)`]: {
--- a/sandpack-react/src/components/common/Stack.tsx
+++ b/sandpack-react/src/components/common/Stack.tsx
@@ -13,7 +13,7 @@ export const stackClassName = css({
   width: "100%",
   position: "relative",
   backgroundColor: "$colors$surface1",
-  transition: "flex $transitions$default",
+  // transition: "flex $transitions$default",
   gap: 1, // border between components
 
   [`&:has(.${THEME_PREFIX}-stack)`]: {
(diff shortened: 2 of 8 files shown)


sandpack-react/src/components/common/Layout.tsx
```diff
@@ -42,10 +41,13 @@ export const layoutClassName = css({
     },
 
     "@media screen and (max-width: 768px)": {
-      height: "auto",
+      [`&:not(.${THEME_PREFIX}-preview, .${THEME_PREFIX}-editor, .${THEME_PREFIX}-preset-column)`]:
```
danilowoz, line 44: Keep min-height for small screen devices

sandpack-react/src/components/common/Stack.tsx
```diff
@@ -13,7 +13,6 @@ export const stackClassName = css({
   width: "100%",
   position: "relative",
   backgroundColor: "$colors$surface1",
-  transition: "flex $transitions$default",
```
danilowoz, line 16: Removed it, otherwise the drag transition looks off

sandpack-react/src/presets/Sandpack.tsx
```diff
@@ -28,20 +28,33 @@ import { classNames } from "../utils/classNames";
...
+export const Sandpack: SandpackInternal = ({
+  options,
+  template,
+  customSetup,
+  files,
+  theme,
+  ...props
+}) => {
+  // fallback values
+  options ??= {};
```
danilowoz, line 40: This reduces the bundler quite a bit

sandpack-react/src/types.ts
```diff
@@ -108,6 +108,7 @@ export interface SandpackOptions {
   showConsole?: boolean;
   closableTabs?: boolean;
   wrapContent?: boolean;
+  resizablePanel?: boolean;
```
danilowoz, line 111: new prop name, which defaults to `true`

alexnm: should resizable=true be the default behavior?

sandpack-react/src/presets/Sandpack.tsx
```diff
@@ -51,122 +64,198 @@ export const Sandpack: SandpackInternal = (props) => {
...
+    const boundaries = Math.min(Math.max(offset, 25), 75);
+
+    if (isHorizontal) {
+      setHorizontalSize(boundaries);
+    } else {
+      setVerticalSize(boundaries);
+    }
+
+    container.querySelectorAll("iframe").forEach((frame) => {
+      frame.style.pointerEvents = "none";
```
alexnm, line 154: can't this be added directly on the preview component? to avoid the dom query here

alexnm: like sandpack state could know a dragging is in progress and propagate that to all components.

sandpack-react/src/presets/Sandpack.tsx
```diff
@@ -194,6 +283,32 @@ const ConsoleCounterButton: React.FC<{
...
+      vertical: {
+        right: 0,
+        left: 0,
+        height: 10,
+        cursor: "ns-resize",
+      },
+      horizontal: {
+        top: 0,
+        bottom: 0,
+        width: 10,
```
alexnm, line 301: did you increase this from the time you shared the demo the first time? I was mentioning that the interaction area with the drag handler was very small

alexnm (approved):
💯

sandpack-react/src/presets/Sandpack.tsx
danilowoz: Yeah, I did. So the drag handler should be more accessible now

sandpack-react/src/presets/Sandpack.tsx
danilowoz: In my opinion, the other components should be stateless and context-less as much as possible, plus, this resizable feature only works on Sandpack preset component in order to keep it isolated from the rest of the components.

PR #711, opened 2023-02-08
Allow clipboard write inside iframe.

nicksrandall:
This change allows application to write to clipboard inside of sandpack iframe.


danilowoz:
Thank you!

PR #720, opened 2023-02-17
fix(react-vite template): use automatic runtime for jsx

jeetiss:
I implemented automatic runtime for VITE-REACT template.

<details>
  <summary> 🤯🤯🤯 </summary>
  This is magic, it works in the browser
</details>

### DIFF
--- a/sandpack-react/src/templates/node/vite-react.ts
+++ b/sandpack-react/src/templates/node/vite-react.ts
@@ -4,9 +4,7 @@ export const VITE_REACT_TEMPLATE = {
   files: {
     ...commonFiles,
     "/App.jsx": {
-      code: `import React from "react"
-
-export default function App() {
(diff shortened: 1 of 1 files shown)


sandpack-react/src/templates/node/vite-react.ts
```diff
@@ -55,11 +52,22 @@ root.render(
           "react-dom": "^18.2.0",
         },
         devDependencies: {
+          "@vitejs/plugin-react": "3.1.0",
           vite: "4.0.0",
           "esbuild-wasm": "0.15.12",
         },
       }),
     },
+    "/vite.config.js": {
```
danilowoz, line 61: React plugin auto inject `React` on every jsx file? Interesting!

danilowoz (approved):
Amazing, thank you! Just left a question

PR #738, opened 2023-02-20
fix(sandpack-react): use automatic runtime for jsx

jeetiss:

Enable `jsx-runtime` for `sandpack-react` package

### DIFF
--- a/sandpack-react/babel.config.js
+++ b/sandpack-react/babel.config.js
@@ -1,7 +1,7 @@
 module.exports = {
   presets: [
     ["@babel/preset-env", { targets: { node: "current" } }],
-    "@babel/preset-react",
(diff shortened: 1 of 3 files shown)


danilowoz:
Hey! What are the advantages of using it? Plus, will this impact the final bundler and how people consume this package?

jeetiss:
@danilowoz sorry i missed the question.

> What are the advantages of using it?

- it should work a little bit faster
- import React is not required in `tsx` files
- typescript doesn't use default import from react

<details>
  <summary>imports </summary>

  before:
Снимок экрана 2023-02-23 в 15 24 34

  after:
Снимок экрана 2023-02-23 в 15 25 05

</details>


storybook and eslint are still configurated in old way, so there is space to improve 

> will this impact the final bundler and how people consume this package?

Nope, it will work as before

PR #741, opened 2023-02-20
fix(client): enable code-splitting and dynamic imports in esbuild

danilowoz:
All kudos to @jeetiss

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "@typescript-eslint/parser": "^4.0.0",
     "babel-eslint": "^10.0.0",
     "babel-jest": "^27.4.5",
-    "esbuild": "^0.12.21",
+    "esbuild": "^0.17.9",
     "eslint": "^7.5.0",
     "eslint-config-prettier": "^8.1.0",
     "eslint-config-react-app": "^6.0.0",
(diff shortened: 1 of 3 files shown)


jeetiss:
this PR fixes only `esm` entries so `commonjs` still can cause bugs and code duplication because `@codesandbox/sandpack-client/clients/node` and `@codesandbox/sandpack-client/clients/runtime` is bundled into `@codesandbox/sandpack-client`

```js
const { loadSandpackClient } = require("@codesandbox/sandpack-client");
const { SandpackRuntime } = require("@codesandbox/sandpack-client/clients/runtime");

(async () => {
  const client = await loadSandpackClient("div > iframe", {});
  console.log(client instanceof SandpackRuntime);
  // ^ will be false for commonjs but should be true
})();
```

[sandbox](https://codesandbox.io/s/sandpack-forked-06pw6f?file=/src/App.jsx)

danilowoz:
Well, `esbuild` says "Splitting currently only works with the "esm" format"
https://github.com/codesandbox/sandpack/pull/753

jeetiss:
Yes, this is impossible to fix with `esbuild`. it is a tool limitation

PR #747, opened 2023-02-21
fix(sandpack-client): move console-feed to dev deps

jeetiss:
move `console-feed` to dev deps to fix warnings about peer dependencies, like this:

`warning "@codesandbox/sandpack-client > console-feed > linkifyjs@2.1.9" has unmet peer dependency "jquery@>= 1.11.0".`

This is safe because `console-feed` is bundled into a string, so end users don't need to install it

To test install `@codesandbox/sandpack-client` with `yarn`

### DIFF
--- a/sandpack-client/package.json
+++ b/sandpack-client/package.json
@@ -47,13 +47,13 @@
   ],
   "dependencies": {
     "@codesandbox/nodebox": "0.1.0",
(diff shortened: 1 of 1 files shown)


danilowoz:
Makes sense, thank you!

PR #758, opened 2023-02-23
fix(sandpack-client): setup build with rollup

jeetiss:
I managed to repeat the build with rollup and it should fix the problem with code duplication

other improvements:

- `consoleHook` is smaller (terser with rollup `16393` vs esbuild `17253`)
- i migrate from the strategy "polyfill node globals" to direct polyfill usage


sandpack-client/src/clients/node/index.ts
```diff
@@ -1,5 +1,7 @@
 /* eslint-disable no-console,@typescript-eslint/no-explicit-any,prefer-rest-params,@typescript-eslint/explicit-module-boundary-types */
 
+import { Buffer } from "buffer";
```
jeetiss, line 3: @danilowoz i'm not super familiar with the code base, could you approve that `buffer` is only one dependency that was polyfill with `@esbuild-plugins/node-globals-polyfill`? i do not see other things

sandpack-client/package.json
```diff
@@ -31,7 +35,7 @@
     "prebuild": "yarn run clean",
     "test": "jest .",
     "lint": "eslint '**/*.ts?(x)' --fix",
-    "build": "node build.js && yarn run build:types",
+    "build": "rollup -c --bundleConfigAsCjs && yarn run build:types",
```
danilowoz, line 38: Maybe we can let rollup generate the types and remove the `build:types`?

sandpack-react/rollup.config.js
```diff
@@ -25,11 +25,14 @@ const configBase = {
 
   plugins: [
     typescript({ tsconfig: "./tsconfig.json" }),
-    replace({ "process.env.TEST_ENV": "false" }),
+    replace({
+      preventAssignment: true,
```
danilowoz, line 29: Why this change is relevant for this PR?

sandpack-client/src/clients/node/index.ts
danilowoz: I believe so

danilowoz (commented):
Could you please update your branch with `main` branch? I introduced important changes there.

sandpack-client/package.json
jeetiss: Yes, it should work

sandpack-react/rollup.config.js
jeetiss: i migrated `rollup-plugin-replace` to `@rollup/plugin-replace` in monorepo

`preventAssignment` flag is new in this plugin, it doesn't replace left values in the assignment to prevent bugs.

danilowoz (approved):
Let's get this merged! Thank you

Issue #769, opened 2023-02-26
Accept Uint8Array as file code input


izznatsir:
# Tell us how you think we can improve Sandpack

## Packages affected

-   [x] sandpack-client
-   [x] sandpack-react

## What is this feature?

`nodebox` accepts both string and Uint8Array to initialize its file system. `sandpack-client` and `sandpack-react` change their API to support it too.

This is useful to make sandpack as closely as possible to local development setup. For example to include static files in public directroy directly inside nodebox such as gltf object for three.js.

## How would your idea work?

Pass in files to `SandpackProvider` in either string or Uint8Array format.

## Do you have any examples of how you would like to see us implement it?

Change the types and pass the files as is for `nodebox` runtime and either convert Uint8Array to utf-8 string or throw error for non-`nodebox` runtime.

izznatsir:
I'll try to submit a PR.

danilowoz:
This is a valid point, and I would love to have it on Sandpack. I have added your feature request to our internal backlog, which we use to prioritize what to build next. We can't make any promises about the timing, but it's on the list! If you'd like to work on this feature with us, please create a pull request, and we'll help you along the way.

PR #773, opened 2023-02-28
feat: Accept Uint8Array as file code alongside string

izznatsir:
- adjust SandpackBundlerFile and SandpackFile types to accept Uin8Array
- create `codeToString` utils to convert Uint8Array into utf-8 string using `TextDecoder`

Feature implementation for #769 .

#769 

Sandpack only acceptsfile code in string format even though nodebox accept both string and Uin8Array.

## What is the new behavior?

Sandpack accepts both string and Uin8Array as file code.

Not yet tested.

## Checklist

- [ ] Documentation;
- [ ] Tests;
- [ ] Ready to be merged;

### DIFF
--- a/sandpack-client/src/utils.ts
+++ b/sandpack-client/src/utils.ts
@@ -17,6 +17,15 @@ export function nullthrows<T>(value?: T | null, err = "Value is nullish"): T {
   return value;
 }
 
+export function codeToString(code: string | Uint8Array): string {
+  if (typeof code === "string") {
+    return code;
+  } else {
+    const decoder = new TextDecoder();
+    return decoder.decode(code);
+  }
+}
+
 const DEPENDENCY_ERROR_MESSAGE = `"dependencies" was not specified - provide either a package.json or a "dependencies" value`;
 const ENTRY_ERROR_MESSAGE = `"entry" was not specified - provide either a package.json with the "main" field or an "entry" value`;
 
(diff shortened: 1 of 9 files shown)


sandpack-client/src/utils.ts
```diff
@@ -17,6 +17,15 @@ export function nullthrows<T>(value?: T | null, err = "Value is nullish"): T {
   return value;
 }
 
+export function codeToString(code: string | Uint8Array): string {
```
izznatsir, line 20: There is also `readBuffer` function in node runtime utils that do the same thing but using `Buffer.from(Uint8Array).toString('utf-8')` instead of `TextDecoder`. Should I remove this and update `readBuffer` implementation to use `TextDecoder` so it can be used outside nodebox?

izznatsir:
SandpackFile and SandpackBundlerFile has no information about what encoding is suitable to encode or decode the code.  So, all are treated as utf-8. Not all files are encodable into utf-8, but TextDecoder does not understand 'binary' string encoding that is available in Node environment. So, maybe we should add another field in SandpackFile and SanpackBundlerFile called `encoding` that accepts either `binary` or `utf-8`, default to `utf-8`.

The `encoding` field can be used whether a file is viewable, so `useActiveCode` util will return undefined if the encoding is  `binary` and code viewer can display a message that the file is not viewable. This is not implemented yet, what do you think?

DeMoorJasper:
Not sure if encoding is really neccessary here, the only runtime that can handle binary data right now is nodebox and there we just write the binary data to the filesystem directly, the only time we need encoding is for reading the pkg.json in sandpack. We can also use something like this if we wanna detect binary for image previews or something: https://github.com/alessioalex/is-binary

Issue #713, opened 2023-02-12
Support  remove default style on client side


Jokcy:
# Tell us how you think we can improve Sandpack

Currently sandpack provided a `SANDPACK_BARE_COMPONENTS` env variable to remove default style generated by stitches, but the syntax seems not work in most cases.

```js
const getNodeProcess = (): false | string | undefined => {
  if (typeof process !== "undefined") {
    return process.env.SANDPACK_BARE_COMPONENTS;
  }

  return false;
};
```

for most client cases `typeof process` will be `undefined` because build tools like webpack will directly replace `process.env.xxx` instead of expose a `process` object to global. So this will always not work on client side. 

Besides the name of the function `getNodeProcess` seems designed for only work on node process?

## Packages affected

-   [ ] sandpack-client
-   [x] sandpack-react

## What is this feature?

Support remove default style on client side.

## How would your idea work?

Just use 

```js
const getNodeProcess = (): false | string | undefined => {
    return process.env.SANDPACK_BARE_COMPONENTS;
};
```

And in case to support some meta framework like next, maybe we can use:

```js
const getNodeProcess = (): false | string | undefined => {
    return process.env.SANDPACK_BARE_COMPONENTS || process.env.NEXT_PUBLIC_SANDPACK_BARE_COMPONENTS  ;
};
```

## Do you have any examples of how you would like to see us implement it?

```js
const getNodeProcess = (): false | string | undefined => {
    return process.env.SANDPACK_BARE_COMPONENTS || process.env.NEXT_PUBLIC_SANDPACK_BARE_COMPONENTS  ;
};
```

Jokcy:
This env variable should work at build time so it should be fine with Nextjs🤔, but if I disable SSR for some page it still not work. Need to look deep to find the reason.

Jokcy:
https://codesandbox.io/p/sandbox/amazing-hooks-qui6b8?file=%2Fpages%2Findex.tsx&selection=%5B%7B%22endColumn%22%3A49%2C%22endLineNumber%22%3A30%2C%22startColumn%22%3A49%2C%22startLineNumber%22%3A30%7D%5D

As this demo shows, nextjs will build code into to version, one for client side and one for server side, the server side build work fine with the env variable but the client build not.

PR #787, opened 2023-03-01
fix: make disable default style env variable work for client side bundler and nextjs

Jokcy:

Make `SANDPACK_BARE_COMPONENTS` work for client bundler, and add Nextjs support.

close https://github.com/codesandbox/sandpack/issues/713

Not work for nextjs

## What is the new behavior?

work for nextjs

## Checklist

- [x] Documentation;
- [ ] Tests;
- [ ] Ready to be merged;

### DIFF
--- a/website/docs/src/pages/getting-started/layout.mdx
+++ b/website/docs/src/pages/getting-started/layout.mdx
@@ -34,7 +34,7 @@ This pattern is compatible with most modern styling systems, including Tailwind,
 
 `@codesanbdox/sandpack-react` relies on [@stitches/core](https://github.com/stitchesjs/stitches) to style its component, which is almost zero-runtime CSS-in-JS framework. However, if you want to get rid of any runtime script or create your own style on top of Sandpack components, we provide a way to return bare components, which will eliminate all Sandpack CSS style.
 
-To do it, you need to pass `SANDPACK_BARE_COMPONENTS` environment variable as `true`, which will remove the Stitches dependency, its execution and return only the HTML of the components.
+To do it, you need to pass `SANDPACK_BARE_COMPONENTS` environment variable (`NEXT_PUBLIC_SANDPACK_BARE_COMPONENTS` for Nextjs) as `true`, which will remove the Stitches dependency, its execution and return only the HTML of the components.
 
 ## Themes
 


danilowoz:
Hey, @Jokcy! Thanks for opening this PR. But I didn't get why we need to make these changes to work on Nextjs
I just created a nextjs sandbox that works as expected on server/client: https://codesandbox.io/p/sandbox/infallible-ardinghelli-ybxh7

Could you highlight what's the current issue and how it solves it?

danilowoz:
@Jokcy I updated your sandbox using another approach and it worked as expected, check the `next.config.js` file
https://codesandbox.io/p/sandbox/restless-sun-3kzd1q

Jokcy:
image

Can not open it, and why I try to sign in with Github it redirect to localhost:4000

Jokcy:
It will work if you use `env` in next.config.js  because env variables provided by this way will both take effect on server/client side. This did solve this problem but it's not the recommended way to provide env variables, it was hard coded into config file and diffcult to make a change in different environment (local/test/production etc). But for this case in nextjs it's enough for me.

Also I'm not sure if the `typeof process !== 'undefined'` will work correctly in other build tools (in my experience it should always be true for client side bundler), I will try vite/webpack later.

Jokcy:
@danilowoz I tried with vite, define variable in vite.config.js

```js
export default defineConfig({
    plugins: [react()],
    define: {
        "process.env.SANDPACK_BARE_COMPONENTS": "true",
    },
});

```
It worked at dev mode because on dev mode vite will actually add an `process` object, but when you bundle the code for production, vite will directly replace the `process.env.xxx` and will not expose a `process` object.

image

So not matter if we add `NEXT_PUBLIC_SANDPACK_BARE_COMPONENTS`, we should still remove the `typeof process !== 'undefined'` condition。

danilowoz:
Got it! It makes sense. Could you please update the documentation with the new env var you introduced? And then, let's get this merged

Jokcy:
@danilowoz done!

danilowoz (approved):
Thank you

origami-z:
This breaks on vite react now with `process is not defined`

image


https://codesandbox.io/p/sandbox/vite-react-sandpack-tqef1u?file=%2FREADME.md

PR #795, opened 2023-03-03
fix(getNodeProcess): check if process is available

danilowoz:
Fix https://github.com/codesandbox/sandpack/pull/787#issuecomment-1452251178

### DIFF
--- a/sandpack-react/src/styles/index.ts
+++ b/sandpack-react/src/styles/index.ts
@@ -12,6 +12,8 @@ import { createStitchesMock } from "./stitches-mock";
 export const THEME_PREFIX = "sp";
 
 const getNodeProcess = (): false | string | undefined => {
+  if (typeof process === "undefined") return false;
+
   return (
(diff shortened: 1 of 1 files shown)


Jokcy:
@danilowoz This breaks the `SANDPACK_BARE_COMPONENTS` ability for disabling default css solution. We need to figure out a better way for this feature, it's not a good idea to depend on `process.env` for client packages, different bundler treat this in default way. 

For example vite recommend we use `import.mate.env` which is supported out box, but if we want to use `process.env` we need additional configuration. So `process.env` is not  a common standard for client packages.

I guess it's better we directly build two version dist files and make `SANDPACK_BARE_COMPONENTS` take effect at build time. The default build will include the default styles and a new build maybe called `styleless/index.js` exculde it.

In this way we can complete exclude `stitches` related code in the styleless build, in the current solution, although we can disable the default style but the code still being bundled into the application which is a waste. And because of the `typeof process` condition tree shaking will take effect to reduce the useless code.

PR #798, opened 2023-03-04
fix: add new build for styleless

Jokcy:

We used try to remove the `typeof process` condition for default styles, but it bring a new issue for vite and a new PR https://github.com/codesandbox/sandpack/pull/795 add the condition back.

We need to understand that this kind of assert will always be `undefinde` for most of bundlers, vite/webpack etc, because there is no `process` in browser and all of them chose to directly replace `process.env.xxx` to the specify value, this means:

```js
const a = process.env.xxx // source code

// after compile

const a = true // the value you specified on config
```

So without remove the `typeof process` condition the _disable default style_ feature will never work on production (for some reason most of bundler actually provide a `process` object on dev mode).

It's not recommeded for client module to rely on `procee.env.xxx` env variables because it's not standard. Different bundle tools handle env with different ways, for example in vite the default supported way to get env is through `import.meta.env.xxx`, so there is no `process.env` unless you config it at `vite.config.js`.

Disable default style will never work on production build.

## What is the new behavior?

Add a new build to remove the default style, the remove action happend at build time which we can control. And we can do more with this to remove all the `css({...})` code in this build to even reduce the final bundle size.

Build with stitches
image

Build without stitches
image

## Checklist

- [ ] Documentation;
- [ ] Tests;
- [ ] Ready to be merged;

### DIFF
--- a/examples/nextjs/pages/index.tsx
+++ b/examples/nextjs/pages/index.tsx
@@ -1,4 +1,4 @@
-import { Sandpack } from "@codesandbox/sandpack-react/unstyled";
+import { Sandpack } from "@codesandbox/sandpack-react";
 import { githubLight, sandpackDark } from "@codesandbox/sandpack-themes";
 
 const Home: React.FC = () => {
--- a/examples/nextjs/pages/index.tsx
+++ b/examples/nextjs/pages/index.tsx
@@ -1,4 +1,4 @@
-import { Sandpack } from "@codesandbox/sandpack-react";
+import { Sandpack } from "@codesandbox/sandpack-react/unstyled";
 import { githubLight, sandpackDark } from "@codesandbox/sandpack-themes";
 
 const Home: React.FC = () => {
--- a/sandpack-react/scripts/rollup-remove-css-transformer.js
+++ b/sandpack-react/scripts/rollup-remove-css-transformer.js
@@ -1,0 +1,51 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+const path = require("path");
+
+const walk = require("acorn-walk");
+const { generate } = require("astring");
+
+module.exports = function () {
+  return {
+    name: "remove-css",
+    transform(code, id) {
+      const ast = this.parse(code);
+
+      const idArr = id.split("/");
+      idArr.pop();
+
+      const sourceDirPath = idArr.join("/");
+
+      let isCssImported = false;
+
+      walk.simple(ast, {
+        ImportDeclaration: (_node) => {
+          if (
+            path.resolve(sourceDirPath, _node.source.value) ===
+            path.join(__dirname, "../src/styles")
+          ) {
+            const specify = _node.specifiers?.find(
+              (specify) => specify.imported.name === "css"
+            );
+
+            if (specify) {
+              isCssImported = true;
+              specify.imported.name = "fakeCss";
+              specify.local.name = "fakeCss";
+            }
+          }
+        },
+        CallExpression: (_node) => {
+          if (_node.callee.name === "css" && isCssImported) {
+            _node.type = "Identifier";
+            _node.name = "fakeCss";
+            delete _node.arguments;
+          }
+        },
+      });
+
+      return {
+        code: generate(ast),
+      };
+    },
+  };
+};
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,6 @@ dist
 
 # Mac
 .DS_Store
+
+# Vscode
+.vscode
--- a/examples/nextjs/tsconfig.json
+++ b/examples/nextjs/tsconfig.json
@@ -9,7 +9,7 @@
     "noEmit": true,
     "esModuleInterop": true,
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "node16",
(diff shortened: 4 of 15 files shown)


danilowoz:
Hey, @Jokcy! I loved this solution!

Could you update the documentation and create a sandbox showing how to use it?

Jokcy:
@danilowoz That's great! I did want to do the same thing but I'm not sure if you will agree this solution so I just implement it in minimal effect.  Great job anyway.

I also want to remove the default css code for the unstyled bundle, for example:
```js
// in Stack.tsx

export const stackClassName = css({
  display: "flex",
  flexDirection: "column",
  width: "100%",
  position: "relative",
  backgroundColor: "$colors$surface1",
  gap: 1, // border between components

  [`&:has(.${THEME_PREFIX}-stack)`]: {
    backgroundColor: "$colors$surface2",
  },
});

```

These code is useless for unstyled bundle, I want to figure out a easy way to remove all these code without a lot of change of our codebase.

Do you want this being include in this PR or another PR should be created?

danilowoz:
Thanks! Yeah, let's try to remove these styles on this PR, feel free to give it a try.

Jokcy:
@danilowoz I added an rollup plugin to remove the css code for unstyled bundle, please review it and give some feedback. It did reduce the size after add this plugin:

image

I also added the filesize plugin to help show file size.

sandpack-react/scripts/rollup-remove-css-transformer.js
```diff
@@ -0,0 +1,51 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
danilowoz, line 1: Absolutely amazing work! Congrats

examples/nextjs/pages/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { Sandpack } from "@codesandbox/sandpack-react";
+import { Sandpack } from "@codesandbox/sandpack-react/unstyled";
```
danilowoz, line 1: Unfortunately, types are still failing here. @Jokcy would you like to give it a try?

Jokcy: Sure, did not notice the types, will check later,

Jokcy: @danilowoz Based on https://github.com/microsoft/TypeScript/issues/50794, set `moduleResolution` to `node16` will fix the types.

sandpack-react/scripts/rollup-remove-css-transformer.js
Jokcy: THX!

examples/nextjs/pages/index.tsx
danilowoz: Great! I spent quite some time on this issue, and I couldn't find a solution. Cool!

danilowoz (approved):
THIS IS GREAT! 🚀

I was wondering if, in the future, we can extract the CSS to static files. What do you think?

Jokcy:
> I was wondering if, in the future, we can extract the CSS to static files. What do you think?

🤔 You mean like pre compile stitches to css files?

Jokcy:
@danilowoz If we generate static css from stitches, I guess it may break the `theme` props usage?

danilowoz:
yeah, it's not an easy issue to solve. 

BTW, can you update this branch with main?

Jokcy:
> BTW, can you update this branch with main?

You mean rebase main to this branch?

Jokcy:
@danilowoz done.

danilowoz:
Hey @Jokcy. It seems that this PR introduces a regression on the Console component. I'm getting the following error only on the documentation (dev-mode works fine), which made me conclude that the bundler has been broken.

I spent quite some time trying to figure out what was going on, but with no success. Maybe you have some idea.
Screenshot 2023-03-12 at 14 09 24

Jokcy:
@danilowoz Got it, will check later. Do we have some unit/e2e tests to help us more confident about the changes? The unstyled bundle should pass all the cases.

danilowoz:
No, we don't. But I'm not if a unit test could help in this case because it looks a like dependency or a bundler issue

Jokcy:
@danilowoz The reason is `import Ansi from "ansi-to-react";` result to `{__esModule: true, default: ƒ}` instead of just get the `default` Ansi component. It's kind of weird...

Jokcy:
@danilowoz Find the reason, it's because we output our es bundle to `mjs`, which seems being handled differently in typescript, change file to `index.js` and `"import": "./dist/index.js"` in packagejson fix this issue. Don't why though, need to do some research.

jeetiss:
I fulgurate out and this is how webpack works. it doesn't apply export default interop to `mjs` files, breaking the "ansi-to-react" dependency. this is a modern module resolution algorithm from the nodejs and it is enabled only for `mjs` files

> The easier way is to output filename like index.esm.js.

so a name with `js` extension will solve the problem.

here is repo with minimal reproduction https://github.com/jeetiss/mjs-commonjs-default

danilowoz (approved):
Thank you @Jokcy! Thank you @jeetiss!

Let's get this merged

Issue #712, opened 2023-02-09
React DevTools component is not available


yyildiz:
# Bug report

## Packages affected

-   [ ] sandpack-client
-   [x] sandpack-react

## Description of the problem
The ReactDevTools component is not available in the SandpackReact package anymore.
It's broken on the documentation page as well: https://sandpack.codesandbox.io/docs/advanced-usage/components#reactdevtools
![image](https://user-images.githubusercontent.com/10237552/217883134-141ff589-e6e9-4150-b910-198f822a4958.png)

### Your Environment

| Software                | Name/Version |
| ----------------------- | ------------ |
| Sandpack-client version |              |
| Sandpack-react version  |              |
| Browser                 |              |
| Operating System        |              |

danilowoz:
We're considering dropping support for ReactDevTools, because Sandpack must be a more agnostic library instead a framework-driven library (although it's written on React). Please, feel free to open it anytime for further discussion

Thanks.

PR #833, opened 2023-03-11
fix(sandpack-react): build ReactDevTools component

jeetiss:
Upgrades `rollup.config` to build `ReactDevTools` properly and moves types generations to rollup

fix: #712


sandpack-react/package.json
```diff
@@ -20,8 +20,7 @@
     "prebuild": "yarn run clean",
     "test": "TEST_ENV=true jest . --transformIgnorePatterns \"node_modules/(?!@codemirror)/\" --modulePathIgnorePatterns \"e2e\"",
     "lint": "eslint '**/*.ts?(x)' --fix",
-    "build": "rollup -c && yarn build:types",
-    "build:types": "tsc -p tsconfig.json",
+    "build": "rollup -c --bundleConfigAsCjs",
```
danilowoz, line 23: We're making this very same change on https://github.com/codesandbox/sandpack/pull/798

sandpack-react/src/components/index.ts
```diff
@@ -7,3 +7,4 @@ export * from "./Preview";
 export * from "./TranspiledCode";
 export * from "./Tests";
 export * from "./Console";
+export * from "./ReactDevTools";
```
danilowoz, line 10: `ReactDevTools` is not currently working, that's why it hasn't been exported

danilowoz (commented):
Thank you for your contribution, but I don't think we can merge it (see comments)

sandpack-react/package.json
danilowoz: Feel free to contribute there too

sandpack-react/src/components/index.ts
jeetiss: yes, i miss that)

Issue #865, opened 2023-04-01
useSandpackConsole and useSandpackShellStdout hooks don't respect maxMessageCount option


SSHari:
# Bug report

## Packages affected

-   [ ] sandpack-client
-   [x] sandpack-react

## Description of the problem
Both the `useSandpackConsole` and `useSandpackShellStdout` hooks don't respect the `maxMessageCount` option. The option is passed in the `useEffect` dependency array, but the default `MAX_MESSAGE_COUNT` value is used instead.

## What were you doing when the problem occurred?
I was trying to change the `maxMessageCount` value for the console in a personal project.

### What steps can we take to reproduce the problem?
1. Render the `Editor` and `Console` components
2. Change the `maxMessageCount` value to something small (e.g. 10)
3. Add more than 10 console log statements to the editor
4. Verify that more than 10 messages are stored in the log

### Your Environment

| Software                | Name/Version |
| ----------------------- | ------------ |
| Sandpack-client version |      -        |
| Sandpack-react version  |    2.1.11       |
| Browser                 |      Chrome        |
| Operating System        |      macOS Ventura        |

danilowoz:
Thank you!

PR #866, opened 2023-04-01
fix(SandpackConsole): clear logs based on maxMessageCount option

SSHari:

This fixes a bug with the `Console` component where the `maxMessageCount` option isn't respected.

The `maxMessageCount` option is ignored and instead the default `MAX_MESSAGE_COUNT` value is used when determining if old logs should be cleaned up.

https://github.com/codesandbox/sandpack/issues/865

## What is the new behavior?

The `maxMessageCount` is respected so that logs are cleaned up based on what the user specifies.

1. Rendered an `Editor` and `Console` component
2. Changed the `maxMessageCount` option
3. Tested with the `vanilla` (client) and `node` (server) templates by logging more lines than the `maxMessageCount` option
4. Verified that the logs earlier in the list were cleaned up

## Checklist

- [ ] Documentation; N/A
- [ ] Tests; N/A
- [x] Ready to be merged;

### DIFF
--- a/sandpack-react/src/components/Console/useSandpackConsole.ts
+++ b/sandpack-react/src/components/Console/useSandpackConsole.ts
@@ -67,7 +67,7 @@ export const useSandpackConsole = ({
             }
           );
 
-          while (messages.length > MAX_MESSAGE_COUNT) {
+          while (messages.length > maxMessageCount) {
             messages.shift();
           }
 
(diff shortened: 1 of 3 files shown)


sandpack-react/src/components/Console/useSandpackConsole.ts
```diff
@@ -67,7 +67,7 @@ export const useSandpackConsole = ({
             }
           );
 
-          while (messages.length > MAX_MESSAGE_COUNT) {
+          while (messages.length > maxMessageCount) {
```
danilowoz, line 70: ahhh 🙈🤦‍♀️🤦🤦‍♂️

danilowoz (approved):

Issue #882, opened 2023-04-04
Allow stylesheets/scripts to be secretly included in static HTML file


joshwcomeau:
# Tell us how you think we can improve Sandpack

I'd like to be able to include stylesheets and scripts within the main HTML file, for the static template, _without_ it showing up within the code editor.

## Packages affected

-   [x] sandpack-client
-   [ ] sandpack-react

## What is this feature?

My goal is to use Sandpack as a sort of supercharged Codepen: The user will be presented with one or more files, and they'll automatically be linked. I'd also love to be able to include a CSS reset, and a CSS file that applies fonts and baseline theme styles for all sandpacks in the app.

Right now, I'd have to link those files explicitly in the HTML file, and so the `index.html` would look something like:

```html
<head>
  <link rel="stylesheet" href="/reset.css">
  <link rel="stylesheet" href="/theme.css">
  <link rel="stylesheet" href="/styles.css">
  <script src="/script.js"></script>
</head>

<h1>Actual content here</h1>
```

I'd prefer for these stylesheets/scripts to be applied automatically, so that the user can only see the relevant content in the editor:

```html
<h1>Actual content here</h1>
```

The best-case for me, in terms of implementation, would be if I could specify an arbitrary chunk of HTML to be included in the `<head>`, that way I could control the order of different stylesheets. But really, the most important thing for me is to be able to link the main CSS file with the main HTML file, without having a <link> show up in the editor.

SSHari:
@joshwcomeau Are the changes in https://github.com/codesandbox/sandpack/pull/884 roughly what you're looking for? Here's a Code Sandbox demo to see what it looks like in action: https://codesandbox.io/s/sandpack-hidden-head-tags-f0kf9z.

joshwcomeau:
@SSHari That looks great! That would definitely work for me :)

I was happy to see that your implementation prepends everything to the <head>, so that the explicitly-linked styles.css overwrites it. That's perfect. 👍

joshwcomeau:
One more thing, semi-related: It would be great if the `<!DOCTYPE html>` could be prepended automatically to the HTML file, if it's not present.

I just found out that this has an extremely curious effect on CSS when it's missing. And I don't want each of my playgrounds to start with `<!DOCTYPE html>`.

PR #884, opened 2023-04-04
feat(Static Template): add hidden head tags option

SSHari:

Feature implementation for https://github.com/codesandbox/sandpack/issues/882.

https://github.com/codesandbox/sandpack/issues/882

## What is the new behavior?

This introduces an option to inject hidden `link` and `script` tags into the static template `index.html` file. A new `hiddenHeadTags` option has been added which has a `links` and `scripts` property. Both properties are arrays that allow for an arbitrary number of _ordered_ tags to be included.

I created a Storybook story with the option set for the `<Sandpack>` and `<SandpackProvider>` components. To test the ordering of the tags, you can switch the order of the link objects so that `/hidden-test-1.css` comes second. The background color for the body _should_ go from  `blue` to `red`.

## Checklist

- [ ] Documentation;
- [ ] Tests;
- [ ] Ready to be merged;

This is a rough implementation of the requested feature in https://github.com/codesandbox/sandpack/issues/882. I know that the feature request mentions wanting to define an arbitrary set of HTML to be appended to the `head` tag. If that's the route we'd like to go, I suppose that an arbitrary HTML string could be passed in the options instead.

One thing to note is that this change only affects the `static` template, but the option is technically available for all the templates. It should be easy enough to type-guard the options based on which template is selected, but I held off on doing anything like that until more discussion happens around this.

Also, any change like this would still require a documentation change as well.

### DIFF
--- a/sandpack-client/src/types.ts
+++ b/sandpack-client/src/types.ts
@@ -59,6 +59,14 @@ export interface ClientOptions {
    * to retrieve npm packages from your own npm registry.
    */
   customNpmRegistries?: NpmRegistry[];
+
+  /**
+   * Configuration for hidden `link` and `script` tags.
+   */
+  hiddenHeadTags?: {
+    links?: { rel: string; href: string }[];
+    scripts?: { src: string }[];
+  };
 }
 
 export interface SandboxSetup {
--- a/sandpack-client/src/types.ts
+++ b/sandpack-client/src/types.ts
@@ -59,14 +59,6 @@ export interface ClientOptions {
    * to retrieve npm packages from your own npm registry.
    */
   customNpmRegistries?: NpmRegistry[];
-
-  /**
-   * Configuration for hidden `link` and `script` tags.
-   */
-  hiddenHeadTags?: {
-    links?: { rel: string; href: string }[];
-    scripts?: { src: string }[];
-  };
 }
 
 export interface SandboxSetup {
--- a/sandpack-client/src/clients/static/index.ts
+++ b/sandpack-client/src/clients/static/index.ts
@@ -44,10 +44,11 @@ export class SandpackStatic extends SandpackClient {
         }
         if (filepath.endsWith(".html") || filepath.endsWith(".htm")) {
           try {
+            content = this.injectDocType(content);
             content = this.injectProtocolScript(content);
-            content = this.injectHiddenHeadTags(
+            content = this.injectExternalResources(
               content,
-              options.hiddenHeadTags
+              options.externalResources
             );
           } catch (err) {
             console.error("Runtime injection failed", err);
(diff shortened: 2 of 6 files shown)


DeMoorJasper:
Amazing to see you jump in this so quickly, like Josh mentioned in the issue it would be nice to also automatically add `<!DOCTYPE html>` if it's missing

Feel free to ping me once this PR is ready for a review, eager to get this merged

sandpack-client/src/types.ts
```diff
@@ -59,6 +59,14 @@ export interface ClientOptions {
    * to retrieve npm packages from your own npm registry.
    */
   customNpmRegistries?: NpmRegistry[];
+
+  /**
+   * Configuration for hidden `link` and `script` tags.
+   */
+  hiddenHeadTags?: {
```
danilowoz, line 66: I'm wondering if we can use `externalResources` (which is for the same purpose but works differently on the runtime client), instead of creating a new one that kind of does the same thing. 

Plus, the export button could carry this value (`externalResources`) to the sandbox on CodeSandbox.io, which would make the experience smoother.

danilowoz: You can see how csb.io implements `externalResources` [here](https://github.com/codesandbox/codesandbox-client/blob/8051314f7dfb71ac23a9a5dfc9ccf39f79508728/packages/app/src/sandbox/external-resources.ts#L85) and get some inspiration

danilowoz (approved):
It looks great! Thanks, @SSHari

sandpack-client/src/types.ts
danilowoz: Great! 

> I think the idea of adding the <!DOCTYPE html> tag if it doesn't exist should be easy enough. 

If I understood correctly if a document misses the document tag it might affect the CSS behavior, which is not necessarily true for `html`, `head`, and `body` tags. So, I'd consider adding only what's needed to ensure a standard experience for everyone, meaning adding the doctype html.

> On the note about carrying over the externalResources value to CodeSandbox.io, it looks like that value might not be used when you click the Open Sandbox button. I tried it with the existing External Resources story which uses the runtime client and the tailwind script doesn't seem to carry over when I click the button.

This is correct. It's not properly working now, due to some backend issues, but let's consider that in the future we will use it.

SSHari: @danilowoz I just pushed an updated version with the following changes:
1. Using the `externalResources` prop
2. Injecting the `<!DOCTYPE html>` tag if it doesn't exist

Here's an example with the setup: https://codesandbox.io/s/sandpack-hidden-head-tags-external-resources-example-iwcny8?file=/src/App.js

**Note:** That example also pulls in an external Tailwind CSS file. That should keep things in-line with how `externalResources` works for the Runtime Client.

Let me know what you think.

danilowoz:
Is anything blocking it? Is it ready for review again?

SSHari:
> Is anything blocking it? Is it ready for review again?

I think it should be ready for review 👍

danilowoz (approved):
So amazing to see your contributions, @SSHari!

Thank you so much once again

joshwcomeau:
Just tried this in my course platform and it works **brilliantly**. Thanks so much, @SSHari @danilowoz @DeMoorJasper ❤️

PR #894, opened 2023-04-07
fix(Static Template): Ensure valid HTML formatting for static template

joshwcomeau:

I suppose it's a bug fix, though it's more about making things a bit more flexible!

In #884, @SSHari added an automatic <!DOCTYPE html> insertion. This works great _if_ a full HTML document is passed, but it produces some funky results otherwise.

For example, if we have this input:

```html
<style>
  p { color: red; }
</style>
<p>Hello World!</p>
```

The `getFileContent` method will prepend the DOCTYPE:

```diff
+<!DOCTYPE html>
<style>
  p { color: red; }
</style>
<p>Hello World!</p>
```

…but when the browser actually renders the preview, that doctype is nowhere to be found.

I think the fundamental problem is that we're not giving it a proper HTML document. It's missing the `<html>`, `<head>`, and `<body>` tags. The browser is remarkably good at taking a chunk of unformed HTML and producing a valid document (eg. the `<style>` gets added to a `<head>`, at least in Chrome), but it doesn't know how to handle `<!DOCTYPE html>`.

## What is the new behavior?

I'm using the [DOMParser API](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser) to convert the provided HTML chunk into a valid HTML document. Then, I grab the text content of the `<html>` tag, slap on the `<!DOCTYPE>`, and return it.

1. I was concerned about whether this would have a negative impact on performance. I timed it using `console.time`. For most "realistic" scenarios, it takes about 1ms (using a 2021 MBP on 6x CPU throttle). I tried it with the full HTML of the Wikipedia homepage, and it took 14ms, but TBH with a document that big, each keystroke took ~500ms total, so the extra 14ms wasn't noticeable.
2. To check that the logic was sound, I tested it with a fully-formed HTML document (like the one in the static template storybook), as well as various "HTML chunks" like the one shown above.

I'd be happy to add unit tests, stories, etc. But first I wanted to see what y'all think of this approach.

## Checklist

- [ ] Documentation;
- [ ] Tests;
- [ ] Ready to be merged;

### DIFF
--- a/sandpack-client/src/clients/static/index.ts
+++ b/sandpack-client/src/clients/static/index.ts
@@ -107,7 +107,6 @@ export class SandpackStatic extends SandpackClient {
     content = readBuffer(content);
 
     // Inject script
-
     content =
       insertHtmlAfterRegex(/<head[^<>]*>/g, content, "\n" + contentToInsert) ??
       contentToInsert + "\n" + content;
--- a/sandpack-client/src/clients/static/index.ts
+++ b/sandpack-client/src/clients/static/index.ts
@@ -13,7 +13,7 @@ import { EventEmitter } from "../event-emitter";
 import { fromBundlerFilesToFS } from "../node/client.utils";
 import type { SandpackNodeMessage } from "../node/types";
 
-import { insertHtmlAfterRegex, readBuffer } from "./utils";
+import { insertHtmlAfterRegex, readBuffer, validateHtml } from "./utils";
 
 export class SandpackStatic extends SandpackClient {
   private emitter: EventEmitter;
@@ -44,7 +44,7 @@ export class SandpackStatic extends SandpackClient {
         }
         if (filepath.endsWith(".html") || filepath.endsWith(".htm")) {
           try {
-            content = this.validateHtml(content);
+            content = validateHtml(content);
             content = this.injectProtocolScript(content);
             content = this.injectExternalResources(
               content,
@@ -79,24 +79,6 @@ export class SandpackStatic extends SandpackClient {
         "accelerometer; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; clipboard-write;"
       );
     }
-  }
-
-  private validateHtml(content: FileContent): FileContent {
-    // Make it a string
-    let contentString = readBuffer(content);
-
-    // Check how long DOMParser takes
-    console.time('domparser');
-
-    const domParser = new DOMParser();
-    const doc = domParser.parseFromString(contentString, "text/html");
-    doc.documentElement.setAttribute("lang", "en");
-
-    const html = doc.documentElement.outerHTML;
-
-    console.timeEnd('domparser');
-
-    return `<!DOCTYPE html>\n${html}`;
   }
 
   private injectContentIntoHead(
--- a/sandpack-client/src/clients/static/index.ts
+++ b/sandpack-client/src/clients/static/index.ts
@@ -85,11 +85,16 @@ export class SandpackStatic extends SandpackClient {
     // Make it a string
     let contentString = readBuffer(content);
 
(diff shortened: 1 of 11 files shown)


sandpack-client/src/clients/static/index.ts
```diff
@@ -81,17 +81,22 @@ export class SandpackStatic extends SandpackClient {
...
-    const docTypeRegex = /<!DOCTYPE*>/gi;
-    if (!docTypeRegex.test(contentString)) {
-      contentString = `<!DOCTYPE html>\n${content}`;
-    }
+    // Check how long DOMParser takes
+    console.time('domparser');
+
+    const domParser = new DOMParser();
+    const doc = domParser.parseFromString(contentString, "text/html");
+    doc.documentElement.setAttribute("lang", "en");
```
joshwcomeau, line 93: So, I definitely don't think we should be forcing all Sandpacks to use English as their HTML language, but it occurred to me that right now, there's no way to set the `lang` attribute in a Sandbox that doesn't render an `<html>` tag. This is kind of an issue; for example, the CSS [hyphens](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens) property only works when the `lang` attribute is set to "en" (and possibly other languages in the future).

sandpack-client/src/clients/static/index.ts
```diff
@@ -81,17 +81,22 @@ export class SandpackStatic extends SandpackClient {
...
+
+    const html = doc.documentElement.outerHTML;
 
-    return contentString;
+    console.timeEnd('domparser');
```
joshwcomeau, line 97: I'll obviously remove the `console.time` stuff, but I left it in for now so y'all could see how I was diagnosing the times I mentioned in the PR :)

sandpack-client/src/clients/static/index.ts
```diff
@@ -81,17 +81,22 @@ export class SandpackStatic extends SandpackClient {
...
+    const domParser = new DOMParser();
+    const doc = domParser.parseFromString(contentString, "text/html");
+    doc.documentElement.setAttribute("lang", "en");
+
+    const html = doc.documentElement.outerHTML;
+
+    console.timeEnd('domparser');
 
-    return contentString;
+    return `<!DOCTYPE html>\n${html}`;
```
joshwcomeau, line 99: The `documentElement.outerHTML` property is a string that includes everything within the `<html>` tags, inclusively. But the `<!DOCTYPE>` sits outside the HTML tag. Therefore, we need to add it back in, even if the user included it in their content.

joshwcomeau:
@danilowoz @DeMoorJasper Slightly unrelated question, but how might I use my fork of Sandpack within my course platform?

I tried to point it to my Github fork in `package.json`:

```json
{
  "dependencies": {
    "@codesandbox/sandpack-react": "joshwcomeau/sandpack#52921c8",
}
```

But this appears to have created the strong structure within `node_modules`, and I couldn't figure it out 😬 any pointers greatly appreciated! ❤️

sandpack-client/src/clients/static/index.ts
danilowoz: I think it's fine to default to `en`. Sandpack contains a lot of opinionated defaults, and we just need to provide a way to make it configurable, for example, persisting the html#lang value from the original document. Good insight, anyway!

danilowoz: TIL, `DOMParser`! Very useful API!

joshwcomeau:
> #894 (comment)

Thanks so much for the unit tests! I hope it wasn't too much trouble.

And TIL about CodeSandbox CI! Will grab the csb.dev URL from that CodeSandbox and use it in my own repo to test these changes in my real-world project, thanks for the tip!

Issue #918, opened 2023-04-26
sandpack.runSandpack() doesn't update preview for static template


zachmerrill:
# Bug report

## Packages affected

-   [ ] sandpack-client
-   [x] sandpack-react

## Description of the problem
When using `template="static"` and `autoReload=false` after editing the `index.html` and pressing Run it does not trigger the preview to update. With `autoReload=true` the preview automatically updates successfully.
This issue doesn't occur with other templates I've tested, such as `template="vanilla"`

## What were you doing when the problem occurred?

Trying to use the static template with my own custom run button using the `useSandpack` hook.

```tsx
const { sandpack } = useSandpack();

<button onClick={() => sandpack.runSandpack()}>Run</button>
```
But this does nothing when the template is static. 

### What steps can we take to reproduce the problem?

The following will not automatically reload and will display the Run button. Make edits to the code and try to run, nothing will happen.

```tsx
<SandpackProvider options={{ autoReload: false }} template="static">
  <SandpackCodeEditor />
  <SandpackPreview />
</SandpackProvider>
```

### Your Environment

| Software                | Name/Version |
| ----------------------- | ------------ |
| Sandpack-client version |       2.6.1       |
| Sandpack-react version  |       2.6.1       |
| Browser                 |        Chrome 112.0.5615.137      |
| Operating System        |        macOS Ventura 13.2.1      |

PR #923, opened 2023-04-28
fix(useClient): track all clients and update clients on runSandpack

SSHari:

A proposed solution for https://github.com/codesandbox/sandpack/issues/918. That bug is about how the `runSandpack` function doesn't work for the `static` template. While working through determining the root cause for that problem, I ran into a few other problems around how we register bundlers.

### Preregistered iFrame Object

When we register a bundler it's only put into the `preregisteredIframes` object if the `useClient` status isn't `running`. The problem is that `runSandpack` only reruns clients that are in the `preregisteredIframes` object, which means if you register a client after things have started up, they won't get picked up by the `runSandpack` function.

### Static Template runSandpack function

When creating a `runtime` client, there's a [global listener that calls updateSandbox](https://github.com/codesandbox/sandpack/blob/main/sandpack-client/src/clients/runtime/index.ts#L126) after the client has initialized. This call to `updateSandbox` is how a newly created client displays its view when `runSandpack` is invoked. The reason the `static` template isn't working is that we don't have a step to call `updateSandbox`. I've added this step to the `static` template constructor and things seem to work.

One thing to note is that this means in a situation where you have `autorun` set to `true`, the `updateSandbox` function will initially be called twice (once from the `static` template constructor and once from [the useClient's watchFileChanges effect](https://github.com/codesandbox/sandpack/blob/main/sandpack-react/src/contexts/utils/useClient.ts#L505)). While this isn't ideal, this is currently how the `runtime` client works as well, so there's consistency.

### Leaking Listeners

This one is harder to prove, but if the `useClient` hook is in a `running` state and you _register_ a new bundler, the code immediately calls `createClient` without cleaning up any existing clients for a given client ID. This means any stale listeners will still be present and won't be cleaned up.

### Examples

I made a [Code Sandbox that better outlines the problems mentioned above](https://codesandbox.io/s/aged-star-1jo1si?file=/src/App.tsx). Verifying the leaking listeners is harder to show in a Code Sandbox, but to verify that problem, you can:
1. Copy the code from the Sandbox
2. Create a story in StoryBook with that code (`template: vanilla` and `autorun: true`)
3. Add a `console.log` statement to the [first line of the runtime client global listener](https://github.com/codesandbox/sandpack/blob/main/sandpack-client/src/clients/runtime/index.ts#L102)
4. Every time you press `Register Bundler`, you should notice that the number of console logs grows because old listeners aren't getting cleaned up

Everything that is mentioned above.

## What is the new behavior?

1. `static` template calls `updateSandbox` as part of the constructor (`runSandpack` should work now)
2. Rename `preregisteredIframes` to `registeredIframes` and track _every_ client that's registered. This way when `runSandpack` is called, it'll work for clients that are registered after `useClient` is in a `running` state
3. Update `createClient` so the first step is to clean up any existing clients with the same client id (removes leaking listeners)

I tried the above examples with the code changes. Everything seems to be working as expected. I also added a Storybook story with a similar setup to the Code Sandbox example.

## Checklist

- [ ] Documentation;
- [ ] Tests;
- [x] Ready to be merged;

### DIFF
--- a/sandpack-react/src/contexts/utils/useClient.ts
+++ b/sandpack-react/src/contexts/utils/useClient.ts
@@ -93,7 +93,7 @@ export const useClient: UseClient = (
    */
   const intersectionObserver = useRef<IntersectionObserver | null>(null);
   const lazyAnchorRef = useRef<HTMLDivElement>(null);
-  const preregisteredIframes = useRef<
+  const registeredIframes = useRef<
     Record<
       string,
       { iframe: HTMLIFrameElement; clientPropsOverride?: ClientPropsOverride }
@@ -121,7 +121,13 @@ export const useClient: UseClient = (
       iframe: HTMLIFrameElement,
       clientId: string,
       clientPropsOverride?: ClientPropsOverride
-    ): Promise<SandpackClientType> => {
+    ): Promise<void> => {
+      // Clean up any existing clients that
+      // have been created with the given id
+      if (clients.current[clientId]) {
+        clients.current[clientId].destroy();
+      }
+
       options ??= {};
       customSetup ??= {};
 
@@ -200,7 +206,7 @@ export const useClient: UseClient = (
          */
       });
 
-      return client;
+      clients.current[clientId] = client;
     },
     [filesState.environment, filesState.files, state.reactDevTools]
   );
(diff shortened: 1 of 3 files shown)


zachmerrill:
This is great! Thanks for working on my issue. Haven't tested the leaking listeners but this definitely fixes `runSandpack()` :)

sandpack-react/src/contexts/utils/useClient.ts
```diff
@@ -216,21 +226,11 @@ export const useClient: UseClient = (
 
   const runSandpack = useCallback(async (): Promise<void> => {
     await Promise.all(
-      Object.keys(preregisteredIframes.current).map(async (clientId) => {
```
danilowoz, line 219: Love these changes. It looks way simpler than before

sandpack-react/src/contexts/utils/useClient.ts
```diff
@@ -292,17 +292,15 @@ export const useClient: UseClient = (
       clientId: string,
       clientPropsOverride?: ClientPropsOverride
     ): Promise<void> => {
+      // Store the iframe info so it can be
+      // used later to manually run sandpack
+      registeredIframes.current[clientId] = {
```
danilowoz, line 297: Great!!

danilowoz (approved):
Thank you so much! 

Very descriptive PR with a lot of context and details! Keep doing it!

PR #964, opened 2023-07-06
Add Solid Babel transform dependency

joshpoll:

Bug fix

The current Solid template fails to run correctly:

Screen Shot 2023-07-06 at 11 30 11 AM

```
Something went wrong

/App.tsx: _tmpl$ is not a function (4:2)

  1 | import { Component } from "solid-js";
  2 | 
  3 | const App: Component = () => {
> 4 |   return <h1>Hello world</h1>
        ^
  5 | };
  6 | 
  7 | export default App;
```

## What is the new behavior?

The template should be fixed

I'm not sure how to test this! I've run it in a separate code sandbox and that worked.
It also matches the dependencies in Ryan Carniato's example sandbox: https://codesandbox.io/s/solid-hello-world-9v13b

## Checklist

- [ ] Documentation; N/A
- [ ] Storybook (if applicable); N/A
- [ ] Tests;
- [ ] Ready to be merged;

### DIFF
--- a/sandpack-react/src/templates/runtime/solid.ts
+++ b/sandpack-react/src/templates/runtime/solid.ts
@@ -36,6 +36,7 @@ render(() => <App />, document.getElementById("app"));`,
       code: JSON.stringify({
         dependencies: {
           "solid-js": "1.3.15",
+          "babel-preset-solid": "1.6.16",
         },
         main: "/index.tsx",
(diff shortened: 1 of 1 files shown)


joshpoll:
Seems like there is also a problem with multiple instances of solid being present.

PR #1002, opened 2023-09-13
fix(node): fix issue with undefined startRoute in iframePreviewUrl

danilowoz:

PR #1036, opened 2023-11-15
[node] Fix undefined startRoute option

davidmccabe:

Fixes a bug with Nodebox when the startRoute option is not set, which caused an invalid iframe URL to be generated.

There was a previous attempt to fix this https://github.com/codesandbox/sandpack/pull/1002 but it didn't actually fix it due to a mistake in operator precedence.

The iframe URL has the string "undefined" appended to it.

## What is the new behavior?

The iframe URL is correct.

No longer seeing the invalid URL when loading a Sandpack.

## Checklist

- [N/A ] Documentation;
- [N/A] Storybook (if applicable);
- [] Tests;
- [ ] Ready to be merged;


danilowoz:
Ohh what a silly mistake! Thanks for fixing it 😀

PR #1037, opened 2023-11-15
feat(node): Added resize event

danilowoz:
Hey, @davidmccabe. Here's the resize event implementation. 

Closes https://github.com/Sandpack/nodebox-runtime/issues/54


davidmccabe:
Awesome, thanks for doing this!

PR #1050, opened 2024-01-03
feat: improve system color scheme support and documentation

danilowoz:
This improves the system color scheme support by listening to system color changes, and setting a preferrable color scheme to the sandpack theme context. Plus, it improves the documentation by mentioning that "auto" value will reflects the system color scheme, which can be the default values (light or dark)


alexnm (commented):
👍

PR #1079, opened 2024-02-08
feat: Private NPM guidance update

JamesACS:
Adding clarification about updated private NPM plan availability to Sandpack docs


danilowoz (approved):
Thank you James

PR #1125, opened 2024-04-25
fix: dynamically reconfigure codemirror instead of recreate on rerender

### LINKED ISSUE #1124: Changing decorators recreates Codemirror
## Scenario
Updating the decorators array moves the cursor to the top of the file. 

## Cause
This is because decorators are specified as part of the following dependency array.

https://github.com/codesandbox/sandpack/blob/f394fad25b82c906210cf0e2cc91b6f1128a7162/sandpack-react/src/components/CodeEditor/CodeMirror.tsx#L323

This results in a full re-instantiation of the codemirror instance, which resets state, extensions etc. 

## Possible fix

Instead changes to decorators need to be updated in a way that does not require a full  re-instantiation.

Something like this might work, however i have limited knowledge of codemirror so this just adds new decorators and doesn't clear old ones.

```js
    React.useEffect(() => {
      if (cmView.current && sortedDecorators) {
        // Add new hightlight decorators
        cmView.current.dispatch({
          effects: StateEffect.appendConfig.of([
            highlightDecorators(sortedDecorators),
          ]),
        });
      }
    }, [sortedDecorators]);
```

## Repro

I've created the following story which can be pasted into this file: 
[sandpack/sandpack-react/src/components/CodeEditor/CodeMirror.tsx](https://github.com/codesandbox/sandpack/blob/f394fad25b82c906210cf0e2cc91b6f1128a7162/sandpack-react/src/components/CodeEditor/CodeMirror.stories.tsx)

If you modify the file a new highlight is generated at random. You will notice that the cursor will be reset after every keypress.

```tsx
type Decorators = Array<{ className: string; line: number }>;

stories.add("DecoratorsDynamic", () => {
  const [decorators, setDecorators] = React.useState<Decorators>([
    // { className: "highlight", line: 9 },
  ]);
  const [file, setFile] = React.useState(`const people = [{
  id: 0,
  name: 'Creola Katherine Johnson',
  profession: 'mathematician',
}, {
  id: 1,
  name: 'Mario José Molina-Pasquel Henríquez',
  profession: 'chemist',
}];

export default function List() {
  const [text, setText] = useState("")
  const listItems = people.map(person =>
    <li>{person}</li>
  );
  return <ul>{listItems}</ul>;
}`);

  React.useEffect(() => {
    const lines = file.split("\n");

    setDecorators([
      {
        className: "highlight",
        line: Math.floor(Math.random() * lines.length),
      },
    ]);
  }, [file]);

  return (
    <SandpackProvider>
      <style>
        {`.highlight, .widget {
        background: red;
      }`}
      </style>
      <CodeEditor
        code={file}
        decorators={decorators}
        fileType="jsx"
        initMode="immediate"
        onCodeUpdate={(newCode) => setFile(newCode)}
      />
    </SandpackProvider>
  );
});
```

danieldelcore:
Closes: #1124 

This PR separates codemirror instantiation logic from subsequent prop updates. 

Previously every time a decorator or other dependency is updated, it would trigger a complete recreation of the sandpack instance, resulting in strange behaviour such as the cursor would be repositioned or elements would flash.

To fix this, I've moved all extension configuration into a seperate hook, when rerun will recreate the entire config and dispatch a codemirror `reconfigure` message which safely updates the instance without any side effects. 

I have added a storybook example which illustrates the behaviour.

### DIFF
--- a/sandpack-react/src/components/CodeEditor/CodeMirror.stories.tsx
+++ b/sandpack-react/src/components/CodeEditor/CodeMirror.stories.tsx
@@ -70,7 +70,7 @@ stories.add("ShowLineNumber", () => (
 #header {
   width: @width;
   height: @height;
-}  
+}
 `}
       fileType="less"
       initMode="immediate"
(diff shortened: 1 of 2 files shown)


danieldelcore:
cc: @CompuIves / @danilowoz would really appreciate your 👁️ 👁️ on this one if you have some time 🙏

danilowoz (approved):
It makes sense! Thank you

danieldelcore:
Legend! Thanks so much!

danilowoz:
It seems that something went wrong with the production build: 
Screenshot 2024-04-26 at 13 49 54

I will investigate it soon

PR #1126, opened 2024-04-26
fix(codemirror): make sure extensions & theme is applied when init

danilowoz:
Fix regression introduce in https://github.com/codesandbox/sandpack/pull/1125


alexnm (commented):
👍

Issue #1173, opened 2024-08-13
Editor is not keyboard accessible


shrilakshmishastry:
# Bug report

## Packages affected

- [ x] sandpack-client
- [x] sandpack-react

## Description of the problem
The editor is not keyboard accessible. This is causing the issue of not reaching to the editor via the keyboard to code or edit the code. 

I am using a tab to navigate. Is there any other keyboard shortcut to reach the editor via the keyboard ?

PR #1184, opened 2024-08-23
fix: accessibility issue in  tabs

shrilakshmishastry:
## Issue:
 - FileTabs has given the role `tab`, but the tab [accessibility pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/#:~:text=Accessibility%20Features,-To%20ensure%20people&text=The%20active%20tab%20has%20a,element%20of%20the%20tab%20element.) is not taken care.
 
 Issues noticed:
 - Not highlighting the tab when the file name is focused
 - Not highlighting the focus state of the editor
 - cross icon or close button not keyboard accessible

Fixed:
 - basic tab accessibility pattern : 
     1. on the tab click to focus on the first tab item initially and on the previously focused tab in case it's accessed earlier
     2. [Managing Focus Within Components Using a Roving tabindex](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#keyboardnavigationinsidecomponents)
     3. highlighting the focus on the tab and editor by default outline of the browser when focused
    
  -  on the focus of the tab, close CTA is visible and is keyboard accessible
  
  ## Screen recording
**Tab pattern implementation**
https://github.com/user-attachments/assets/6221d6dd-1050-486c-adb6-9e2e6b1420f2

Editor accessibility
https://github.com/user-attachments/assets/7049baa6-762d-4015-85b6-21fd20a20f41

- Accessibility fix

## Checklist

- [ ] Documentation; N/A
- [ ] Tests;
- [x] Ready to be merged;

This fix partially https://github.com/codesandbox/sandpack/issues/1173

### DIFF
--- a/sandpack-react/package.json
+++ b/sandpack-react/package.json
@@ -39,7 +39,7 @@
     "lint": "eslint '**/*.ts?(x)' --fix",
     "build": "rollup -c",
     "build:publish": "yarn build",
-    "start": "tsc -p tsconfig.esm.json --watch",
+    "start": "tsc -p tsconfig.json --watch",
     "dev": "storybook dev -p 6006 --quiet",
     "typecheck": "tsc",
     "format": "prettier --write '**/*.{ts,tsx,js,jsx}'",
--- a/sandpack-react/package.json
+++ b/sandpack-react/package.json
@@ -39,7 +39,6 @@
     "lint": "eslint '**/*.ts?(x)' --fix",
     "build": "rollup -c",
     "build:publish": "yarn build",
-    "start": "tsc -p tsconfig.json --watch",
     "dev": "storybook dev -p 6006 --quiet",
     "typecheck": "tsc",
     "format": "prettier --write '**/*.{ts,tsx,js,jsx}'",
--- a/sandpack-react/src/components/CodeEditor/index.tsx
+++ b/sandpack-react/src/components/CodeEditor/index.tsx
@@ -98,7 +98,12 @@ export const SandpackCodeEditor = forwardRef<CodeMirrorRef, CodeEditorProps>(
       <SandpackStack className={classNames("editor", [className])} {...props}>
         {shouldShowTabs && <FileTabs closableTabs={closableTabs} />}
 
-        <div className={classNames("code-editor", [editorClassName])}>
+        <div
+          aria-labelledby={`${activeFile}-tab`}
+          className={classNames("code-editor", [editorClassName])}
+          id={`${activeFile}-tab-panel`}
+          role="tabpanel"
+        >
           <CodeMirror
             key={activeFile}
             ref={ref}
--- a/sandpack-react/src/components/CodeEditor/index.tsx
+++ b/sandpack-react/src/components/CodeEditor/index.tsx
@@ -6,6 +6,7 @@ import { useActiveCode } from "../../hooks/useActiveCode";
 import { useSandpack } from "../../hooks/useSandpack";
 import type { CustomLanguage, SandpackInitMode } from "../../types";
 import { useClassNames } from "../../utils/classNames";
+import { useSandpackId } from "../../utils/useAsyncSandpackId";
 import { FileTabs } from "../FileTabs";
 import { RunButton } from "../common/RunButton";
 import { SandpackStack } from "../common/Stack";
@@ -94,14 +95,21 @@ export const SandpackCodeEditor = forwardRef<CodeMirrorRef, CodeEditorProps>(
       updateCode(newCode, shouldUpdatePreview);
     };
 
+    const activeFileUniqueId = useSandpackId();
+
     return (
       <SandpackStack className={classNames("editor", [className])} {...props}>
-        {shouldShowTabs && <FileTabs closableTabs={closableTabs} />}
+        {shouldShowTabs && (
+          <FileTabs
+            activeFileUniqueId={activeFileUniqueId}
+            closableTabs={closableTabs}
+          />
+        )}
 
         <div
-          aria-labelledby={`${activeFile}-tab`}
+          aria-labelledby={`${activeFile}-${activeFileUniqueId}-tab`}
           className={classNames("code-editor", [editorClassName])}
-          id={`${activeFile}-tab-panel`}
+          id={`${activeFile}-${activeFileUniqueId}-tab-panel`}
           role="tabpanel"
         >
           <CodeMirror
--- a/sandpack-react/src/components/CodeEditor/styles.ts
+++ b/sandpack-react/src/components/CodeEditor/styles.ts
@@ -27,9 +27,6 @@ export const editorClassName = css({
   [`.${placeholderClassName}`]: {
     padding: "$space$4 0",
   },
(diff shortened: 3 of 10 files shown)


sandpack-react/package.json
```diff
@@ -39,7 +39,7 @@
     "lint": "eslint '**/*.ts?(x)' --fix",
     "build": "rollup -c",
     "build:publish": "yarn build",
-    "start": "tsc -p tsconfig.esm.json --watch",
+    "start": "tsc -p tsconfig.json --watch",
```
danilowoz, line 42: Interesting! I think this command is not used anywhere and we could even remove it

sandpack-react/src/components/CodeEditor/index.tsx
```diff
@@ -98,7 +98,12 @@ export const SandpackCodeEditor = forwardRef<CodeMirrorRef, CodeEditorProps>(
       <SandpackStack className={classNames("editor", [className])} {...props}>
         {shouldShowTabs && <FileTabs closableTabs={closableTabs} />}
 
-        <div className={classNames("code-editor", [editorClassName])}>
+        <div
+          aria-labelledby={`${activeFile}-tab`}
+          className={classNames("code-editor", [editorClassName])}
+          id={`${activeFile}-tab-panel`}
```
danilowoz, line 104: Considering that a page can contain multiple Sandpack instances with similar content, this ID will not be unique to the page. Is there any reason why you need an ID here? There is an alternative to generate unique identifier inside Sandpack

sandpack-react/src/components/FileTabs/index.tsx
```diff
@@ -25,12 +25,28 @@ const tabsScrollableClassName = css({
...
+
 const closeButtonClassName = css({
   padding: "0 $space$1 0 $space$1",
   borderRadius: "$border$radius",
   marginLeft: "$space$1",
   width: "$space$5",
   visibility: "hidden",
+  cursor: "pointer",
+  position: "absolute",
+  color: "green",
```
danilowoz, line 48: Is this color really intended?

shrilakshmishastry: no, I missed to remove it 😬

shrilakshmishastry:
Hey @danilowoz 👋  Thanks for reviewing !! Did you get a chance to try it out ?  
I am not satisfied with how I handled the outline when the tab and code editor are focused , using the default browser outline. Please suggest how to improve this.

sandpack-react/src/components/CodeEditor/index.tsx
shrilakshmishastry: yes, this ID is used inside the `FileTabs` component, to add the `aria-controls` attribute. https://github.com/codesandbox/sandpack/blob/e571870afcb0bd3ac1382fd81b1a7de2cca79db1/sandpack-react/src/components/FileTabs/index.tsx#L205 

is there a way I can get a unique ID and pass it between `FileTabs` and `CodeEditor`?

One way I am thinking is to add a unique id with activeFile , what's your opinion ?

danilowoz: I just introduced a new hook (`useSandpackId`) to your branch and I think you could use it to generate ids

danilowoz:
I'm a bit hesitant about adding a ring focus over the code editor, giving the following reasons:
- There is already a visual focus indication in the component (active line), and I just improved to remove the style when there is no focus;
- It might disturb the layout once the editor wasn't designed to contain borders except for the border that wrap everything
- If your project really needs a more prominent focus indication, you can always implement it on your side by overwriting the default styles.

Let me know what you think :)

shrilakshmishastry:
> I'm a bit hesitant about adding a ring focus over the code editor, giving the following reasons:
> 
> * There is already a visual focus indication in the component (active line), and I just improved to remove the style when there is no focus;
> * It might disturb the layout once the editor wasn't designed to contain borders except for the border that wrap everything
> * If your project really needs a more prominent focus indication, you can always implement it on your side by overwriting the default styles.
> 
> Let me know what you think :)

Hey 👋 , 
1. I tried focusing the editor using a tab but could not see the focus highlight on both dark and light modes. 
2. focusing is outline right, it's not going to affect any layout , correct me here please 
3. can you please explain to me the last point a bit more?

shrilakshmishastry:
[feat: add unique id to each tab and tab panel for aria-* attributes](https://github.com/codesandbox/sandpack/pull/1184/commits/a127706ac21e44d8eadf032eb235f5e0b34d5410)

1. I used `useSandpackId()` in this commit to generate unique values for the ` aria-labelledby` and `aria-controls` attributes. 

2. added back outline to the editor on focus https://github.com/codesandbox/sandpack/blob/a127706ac21e44d8eadf032eb235f5e0b34d5410/sandpack-react/src/components/CodeEditor/styles.ts#L31

Screenshot 2024-09-01 at 8 55 07 PM

sandpack-react/src/components/CodeEditor/styles.ts
```diff
@@ -27,6 +27,9 @@ export const editorClassName = css({
   [`.${placeholderClassName}`]: {
     padding: "$space$4 0",
   },
+  "&:focus-within": {
+    outline: "$colors$accent auto 1px",
```
danilowoz, line 31: @shrilakshmishastry, this can be implemented on your end, meaning inside the application that implements Sandpack, instead of adding a default style that might not be wanted by everyone. For a basic and generic experience, the editor already provides the necessary hint when it's focused, so I don't think adding such an opinionated style will be a good idea now. 

For example, this can be added to your application:
```css
.sp-code-editor:focus-within  {
  outline: 1px auto red;
}
```

shrilakshmishastry: hmm, got it 👍  make sense, I will revert this

shrilakshmishastry: reverted [chore: remove focus-within added in codeEditor/styles](https://github.com/codesandbox/sandpack/pull/1184/commits/e26006e05ceb5fd10f9dca6157990f48b93a3911)

shrilakshmishastry: image

danilowoz (approved):

danilowoz:
It looks good to me! But could you please check the Typescript job and fix the issues?

shrilakshmishastry:
In the [refactor: make activeFileUniqueId as optional paramater in FileTabs](https://github.com/codesandbox/sandpack/pull/1184/commits/b63d307ebf0d337b3a300cdd4e3ff87e0308488a) commit, 

1. I have made ` activeFileUniqueId` as optional. So it doesn't break the existing consumption 
2.  fixed the ts type issue, it was due to not passing this prop in `codeviewer` component 
3. moved aria-* to div level in FIleTab, this is done  as voice-over was not considering tab  as in a series of tabs when it was added to button.

@danilowoz can you please re-review ?

danilowoz (approved):

PR #1188, opened 2024-09-02
docs: drop nodebox examples

danilowoz:


huxulm:
Does this mean that the original templates are no longer available?

danilowoz:
Hey, all Nodebox templates are still available on Sandpack.

Thanks for asking :)

davekiss:
@danilowoz why were these dropped? looking for some examples for a spike on my end

danilowoz:
Due to a lack of time and resources to address the numerous issues related to Nodebox, we have decided to stop recommending it for now. We will reconsider its recommendation once proper support is available.

lilpolymath:
@danilowoz Since you are "deprecating" `Nodebox`, two things:
- what does that mean for testing since it is also no part of the recommended templates.
- is there any documents that collates what you have, in case the community wants to take it up? I really want to use Nodebox.

PR #1197, opened 2024-09-10
fix(sw): support fixed id value

danilowoz:


sandpack-react/src/Playground.stories.tsx
```diff
@@ -7,13 +7,17 @@ export default {
   title: "Intro/Playground",
 };
 
+localStorage.setItem("SANDPACK_INTERNAL:URL-CONSISTENT-ID", "123123123");
```
CompuIves, line 10: Is this needed? Should the user do this too?

sandpack-react/src/contexts/utils/useClient.ts
```diff
@@ -159,6 +162,26 @@ export const useClient: UseClient = (
         }, timeOut);
       }
 
+      const getStableServiceWorkerId = async () => {
+        const key = `SANDPACK_INTERNAL:URL-CONSISTENT-ID`;
+        const fixedId = localStorage.getItem(key);
+        if (fixedId) {
```
CompuIves, line 168: Maybe we should do `localStorage.setItem` for the id if it turns out it's not set (with a random id). Then we don't have to ask the user to set it.

PR #1295, opened 2026-03-26
fix(client): capture console values at log time

### LINKED ISSUE #1294: Console logs can show mutated values instead of call-time snapshots
### Repro

In a Sandpack preview, run:

```js
const numbers = [1, 2, 3];
console.log(numbers);
numbers.push(4);
console.log(numbers);
```

### Current behavior

The first log can show the mutated value (for example `[1,2,3,4]`) instead of the original snapshot (`[1,2,3]`).

### Expected behavior

Each console message should reflect the value at the moment `console.log(...)` was called.

### Root cause

`@codesandbox/sandpack-client`'s inject script currently relies on `console-feed`'s `Hook(...)`, which defers parsing with `setTimeout(...)`. That delayed parse reads references after mutation, so the first message can be stale/incorrect.

### Suggested direction

Capture/parse/encode the console arguments synchronously in the injected console wrapper before later mutations happen.

Sembauke:
## Summary

Fixes console payload mutation in Sandpack previews by capturing log data at call time.

When a mutable value is logged and then mutated immediately after, the first log could incorrectly show the later mutated value. This happened because the previous hook path relied on delayed parsing.

This PR updates the inject script to:

- wrap console methods directly,
- parse/encode payloads synchronously before mutation can occur,
- preserve the existing postMessage payload shape.

## Tests

Added `sandpack-client/src/inject-scripts/consoleHook.test.ts` with regressions for:

- mutable array snapshot behavior,
- unchanged primitive payload behavior.

## Verification

- `corepack yarn workspace @codesandbox/sandpack-client test src/inject-scripts/consoleHook.test.ts`
- `corepack yarn workspace @codesandbox/sandpack-client build`

Closes #1294

### DIFF
--- a/sandpack-client/src/inject-scripts/consoleHook.test.ts
+++ b/sandpack-client/src/inject-scripts/consoleHook.test.ts
@@ -1,0 +1,87 @@
+/**
+ * @jest-environment jsdom
+ */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import Methods from "console-feed/lib/definitions/Methods";
+
+const flush = (): Promise<void> =>
+  new Promise((resolve) => {
+    setTimeout(resolve, 0);
+  });
+
+describe("consoleHook", () => {
+  const originalScope = (globalThis as any).scope;
+  const originalMethods = new Map<string, unknown>();
+
+  beforeEach(() => {
+    jest.resetModules();
+    originalMethods.clear();
+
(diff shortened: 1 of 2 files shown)


s1ddiq:
bro can they push ts :sob:

Sembauke:
This repo has been inactive since last year.  I would not count on it @s1ddiq

PR #1310, opened 2026-09-02
fix(client): secure postMessage communications and close iframe injection vector

vibemasshq-dev:
### What does this PR do?
This PR hardens the `postMessage` communication layer in `sandpack-client/src/clients/node/index.ts` and `inject-scripts/historyListener.ts`. 

While reviewing the event listeners, I noticed a critical nuance in how messages are handled:
1. **Inbound Injection Vector:** Although `messageChannelId` provides a good security token for most dispatched events, the `PREVIEW_LOADED_MESSAGE_TYPE` check in `globalListeners` lacks both channel ID and origin validation. Any malicious page that knows this public event type could send it to the Sandpack instance and trigger `injectScriptToIframe`. This PR scopes that listener strictly to the expected `bundlerURL` origin.
2. **Outbound Data Leakage:** In the `dispatch` method, `urlback` and `urlforward` messages are broadcast to the iframe using a wildcard `*` target origin. If the preview iframe were somehow navigated to an unexpected origin, it would receive these internal navigation commands. This PR scopes those outbound messages specifically to `new URL(this.iframePreviewUrl).origin`.

### Architecture Heads-Up 🏗️
While investigating the messaging layer, I was looking through the `runtime` client folder and noticed that `sandpack-client/src/clients/runtime/index.ts` is growing into quite a monolithic module (~800+ lines handling everything from bundler compilation to iframe management). 

As Sandpack continues to scale, it might be worth decomposing that core runtime client into smaller, more focused sub-modules (e.g., separating the message handlers from the state manager) to improve testability and make onboarding easier for future contributors!

Best,
@vibemasshq-dev

### DIFF
--- a/sandpack-client/src/clients/node/index.ts
+++ b/sandpack-client/src/clients/node/index.ts
@@ -251,7 +251,13 @@ export class SandpackNode extends SandpackClient {
   private async globalListeners(): Promise<void> {
     window.addEventListener("message", (event) => {
       if (event.data.type === PREVIEW_LOADED_MESSAGE_TYPE) {
-        injectScriptToIframe(this.iframe, this.messageChannelId);
+        if (
+          this.options.bundlerURL &&
+          event.origin !== new URL(this.options.bundlerURL).origin
+        ) {
+          return;
+        }
+        injectScriptToIframe(this.iframe, this.messageChannelId, window.location.origin);
       }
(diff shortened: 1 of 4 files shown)


vibemasshq-dev:
Hello, any review on this? Thank you

